PR #68, opened 2015-03-19
Change curator to editor across the board

rfox2:


rfox2:
Jon: good to go, I think

PR #88, opened 2015-04-17
Showcase editor fix

danthewolfe:
Similar fix to the one for DEC-187 on beehive.


jaronkk:
I like the simplified code, but right now it's rendering "false" if there's no description.  It should work if you change `var rawMarkup = false;` to `var rawMarkup;`

PR #95, opened 2015-04-29
Add redector to item description and transcript fields

rfox2:
**What** I added the redactor editor to the item description and transcript fields in the UI

**How** I modified a couple of view partials, one for the item edit page, and one associated with the item list for a collection

**Why** The form output for the item edit page needed to be changed to remove the old type, and add a class that renders the redactor editing tools; the description needed to be changed on the item list so that it output raw HTML instead of using the standard Draper helper method to accomodate HTML in the descriptive field

**Note** Beehive will need to be updated to accommodate for HTML being in the item description and transcript fields (see image):

screen shot 2015-04-29 at 2 43 13 pm

### DIFF
--- a/app/views/items/_form.html.erb
+++ b/app/views/items/_form.html.erb
@@ -3,11 +3,9 @@
 
 <%= f.input :title, as: :string %>
 
-<%= f.input :description, as: :text do %>
-  <%= f.text_area :description, class: 'text optional form-control', rows: '3' %>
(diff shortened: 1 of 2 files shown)


danthewolfe:
:+1: 

@rfox2  can you create a JIRA ticket for adding HTML support to item descirption on Beehive?

jaronkk:
:+1:

PR #96, opened 2015-04-29
DEC-242: Removed etag based caching from all api controllers and updated...

jgondron:
... the specs to test for content based etag generation instead.


jaronkk:
:+1:

danthewolfe:
:+1:

PR #98, opened 2015-04-30
DEC-245: Added collection to etag generation on all of the other control...

jgondron:
...lers. We decided to change users to content based since it does not appear to be directly used atm, and the data is fairly small to render anyway. We can worry about that one if it becomes an issue.


jaronkk:
:+1:

danthewolfe:
:+1:

PR #100, opened 2015-04-30
Add site introduction 

jonhartzler:
Beehive needs to have a separation between the site introduction and the introduction for the exhibit.  

To do this I added a field to the exhibit table, added the field to the exhibit form, and added the field to the api.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20150416184347) do
+ActiveRecord::Schema.define(version: 20150430164752) do
 
   create_table "collection_users", force: :cascade do |t|
     t.integer  "user_id",       limit: 4, null: false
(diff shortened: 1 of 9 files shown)


db/schema.rb
```diff
@@ -44,6 +44,7 @@
     t.string   "image_content_type", limit: 255
     t.integer  "image_file_size",    limit: 4
     t.datetime "image_updated_at"
+    t.text     "short_description",  limit: 65535
```
danthewolfe: Is this too big of a limit?

jaronkk: This is the default limit for a text field.  Rails has started dumping the full limits in the schema.rb.  The only reason I might limit it is if MySQL has to allocate that much space per row, even if the full value isn't used (but I don't know exactly how it stores text)

jonhartzler: In general it seems every time I try and be judicious about the length all I ever get is an opportunity to extend it later.  I would only look at this if it was slow in the query.   That said it is probably too long.

danthewolfe: I kind of figured it was a default value. Can or should we limit the available text length in Redactor? (This is probably an issue that can be handled separately from this PR).

danthewolfe:
:+1:

jaronkk:
:+1:

PR #101, opened 2015-05-01
Add collection subtitle

jonhartzler:
We need a subtitle style field for collection names to display names across 2 lines. 

I updated the database, updated the form and added the field to the api.


jaronkk:
:+1:

jaronkk:
Other than the one hound message, :+1:

PR #103, opened 2015-05-04
Cache key generators for v1

jgondron:
Cache keys used for etag generation are now being handled by CacheKeys::Generator classes. For now I only have the v1 controllers implemented so that we can review how this is being handled. If we like this, then I can do the rest of the controllers.


danthewolfe:
:+1:

PR #105, opened 2015-05-05
Exhibits updated/created dates

jgondron:
This will add updated_at and created_at date fields so that caching will work properly. Please note that it does perform a data update within the migration.


jonhartzler:
:+1:

danthewolfe:
:+1:

PR #106, opened 2015-05-06
Add about text

jonhartzler:
We needed a way for the admin to edit the button on the info text. 

To do this I added the a field to the exhibit class, edited the site editor section of the site to have a text area, and I added the about text to the api.  In addition I refactored the way the site editor works so that we can have the different texts on different pages.   

Also I added curly into the project as an example.  You can see how curly was used in side_nav_presenter.rb, _side_nav.html.erb, and side_nav_presenter_spec.rb.   I believe we can easily remove this test if we decide we really don't like it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,7 +22,9 @@ gem 'hesburgh_api', git: 'https://github.com/ndlib/hesburgh_api.git'
 
 gem 'simple_form', '~> 3.1.0rc1'
 
+gem 'curly-templates'
 gem 'draper'
+
 # used to normaize the characters in a title sort
 gem 'sort_alphabetical'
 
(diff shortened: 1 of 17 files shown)


danthewolfe:
Let's give curly a shot.

:+1:

rfox2:
I love the idea of curly. I'm a little uncomfortable just dabbling in it apart from an intentional decision to begin modeling our our views in this way in a consistent way. Debugging or refactoring in the future might get complicated if we lose consistency, but I really do love the architecture of curly.

:+1:

jonhartzler:
@rfox2 Yeah I choose to put it in because I was about to duplicate a pattern that we have looked at and said is not good enough.  I also figured that it is very simple to fall back on what we have done before and that it would be better to see how it fits in to a specific problem that we have.  I find it completely reasonable to remove the the curly implementation.  I mostly wanted to see how it fit.

PR #108, opened 2015-05-07
update material design to fix errors and be up to date

jonhartzler:
Updated material design because it was behind and because the old version referenced fonts that 404ed.


rfox2:
:+1:

danthewolfe:
:+1:

PR #109, opened 2015-05-07
DEC-235: Convert all single-quotes to double-quotes

jgondron:
Ran rubocop with auto-correct using only the one style-guide: https://github.com/bbatsov/ruby-style-guide#consistent-string-literals. This will reduce our current offenses from 1941 to 108.


rfox2:
I'm :+1: after the hound corrections are made

PR #111, opened 2015-05-11
Dec 276 collections publish unpublish

rfox2:
Sorry, this pull request is a little ugly.

**What** Added an element to the left nav of honeycomb to allow easy publishing / unpublishing of collections

**Why** This was the second half of the work to add functionality to publish / unpublish collections in the application

**How** The principle change is the addition of a React component that interacts with v1 API calls, which were also added as a separate commit.

**NOTE** For Jon Hartzler: I commented out two tests that began failing after one of my commits, and I'm not sure what the purpose of the tests are or what the method being tested is supposed to do. Prior to merge, let's take a look at those tests because at this point I'm not sure why they are broken.


app/queries/collection_query.rb
```diff
@@ -25,6 +25,14 @@ def public_find(id)
     relation.find_by!(unique_id: id, published: true)
   end
 
+  def private_find(id)
```
jaronkk: Do we need the `private_find` method?

jonhartzler:
Ok I have a couple comments.   

1.  I think the publish button looks really good in the menu I was not sure I could imagine how it was going to come out but I like it.  

2.  the bootstrap button style has a font-weight 400 on it seems to me like it is too strong and out of place since nothing else is bolded in the nav.  

3.  If you change the path variables passed into the react component to v1_collection_publish_url(id)  you won't need to specify the full url in the react component.  (which is currently localhost)

app/queries/collection_query.rb
rfox2: We don't. I thought I would need it but then it didn't make sense so I created the `any_find` method instead. I was ambivalent about removing it but I'm not opposed to it. Remove?

jaronkk: Yeah, I'd like the method removed if we don't have an immediate need for it.

danthewolfe:
Address @jonhartzler's concerns and I'm okay with this. :+1:

app/queries/collection_query.rb
rfox2: Done. It took me weeks to write that. :unamused:

jonhartzler:
Also: 

Add these end points to  https://app.apiary.io/honeycomb/editor

PR #112, opened 2015-05-12
DEC-308-Bust the cache on redeploy

jgondron:
On start, the application will now generate a unique key that will be prepended to all cache key strings used to generate etags, causing all etags to change after a restart of the server.


jonhartzler:
:+1:

danthewolfe:
Very helpful. :+1: 

:shipit:

PR #116, opened 2015-05-13
DEC-306: Added tests for next and previous

jgondron:
What: Next/Previous specs for V1::ShowcaseJSONDecorator and ShowcaseQuery
Why: We added methods to get the next and previous showcases and needed specs to test them.


danthewolfe:
:+1:

spec/queries/showcase_query_spec.rb
```diff
@@ -60,4 +60,24 @@
       expect { subject.public_find("asdf") }.to raise_error ActiveRecord::RecordNotFound
     end
   end
+
+  describe "next" do
+    let(:showcase) { instance_double(Showcase, id: 1, title: "title", exhibit_id: 1, order: 1) }
+
+    it "gets the next record based on order" do
+      expect(relation).to receive(:where).with(exhibit_id: showcase.exhibit_id).and_return(relation)
```
rfox2: Both of these sets of expectations actually do nothing. I talked to Justin about this today and he agrees. The mocks completely separate the expectation from the method logic, so we aren't testing anything. This is important enough functionality, though, that I think it should be tested. But, this gets us into the conundrum of whether and how to test Active Record functionality. But my suggestion is that these tests be removed as they stand.

rfox2:
:question: Not sure, but for now I vote :-1: as this stands

jonhartzler:
:+1:

PR #119, opened 2015-05-18
Dec 313

jgondron:
**Why:** We want to start moving these types of data operations to the api. We are starting with this one since we are modifying the save operation to include the metadata fields for an item. This needs to be done in order to continue progress on that ticket (https://jira.library.nd.edu/browse/DEC-273)
**How:** For the most part this was a copy/paste from the other ItemsController, except that here we render the item data only for the json and we also include error information since any services that use this will need this info to deal with the unsuccessful save.


danthewolfe:
:+1:

jonhartzler:
:+1:

PR #120, opened 2015-05-18
Dec 313

jgondron:
**Why:** 
Need to fix two bugs with V1::Item.update:
  1. Cannot find the object by the id given by the api
  1. Has a bug with rendering the partial when the save is successful.

**How:** 
  1. Modified to use public find to find the object since the API objects will always use the public id instead of the internal id
  1. Added an update jbuilder that simply renders the partial after decorating it.


jonhartzler:
:+1:

danthewolfe:
:+1:

PR #123, opened 2015-05-20
Dec 314

jonhartzler:
We needed to start using our api for item updates.  

I added use of an ajax call to the item api and I added display of error messages returned from the server.

### DIFF
--- a/app/assets/javascripts/components/Form.js.jsx
+++ b/app/assets/javascripts/components/Form.js.jsx
@@ -32,8 +32,8 @@ var Form = React.createClass({
   formErrorAlert: function () {
     if (this.props.hasErrors)  {
       return (
-        <div className="alert alert-danger" role="alert">
-          <strong>Opps!!</strong> errors prohibited this form being saved.
+        <div className="alert alert-warning" role="alert">
+          Please complete the highlighted fields in order to continue.
         </div>
       )
     }
--- a/app/assets/javascripts/components/Form.js.jsx
+++ b/app/assets/javascripts/components/Form.js.jsx
@@ -12,6 +12,7 @@ var Form = React.createClass({
     url: React.PropTypes.string.isRequired,
     authenticityToken: React.PropTypes.string.isRequired,
     method: React.PropTypes.string,
+    hasErrors: React.PropTypes.object,
   },
 
   getDefaultProps: function() {
@@ -28,12 +29,24 @@ var Form = React.createClass({
     }
   },
 
+  formErrorAlert: function () {
+    if (this.props.hasErrors)  {
+      return (
+        <div className="alert alert-danger" role="alert">
+          <strong>Opps!!</strong> errors prohibited this form being saved.
+        </div>
+      )
+    }
+    return "";
+  },
+
   render: function () {
     return (
       <form action={this.props.url} method={ this.formMethod() } >
         <input name="utf8" type="hidden" value="✓" />
         <input type="hidden" name="_method" value={this.props.method} />
         <input type="hidden" name="authenticity_token" value={this.props.authenticityToken} />
+        {this.formErrorAlert()}
         {this.props.children}
       </form>
     );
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/components/item_meta_data_form/ItemMetaDataForm.js.jsx
```diff
@@ -30,17 +55,32 @@ var ItemMetaDataForm = React.createClass({
...
+          <StringField objectType="item" name="title" required="true" title="Title" value={this.state.formValues['title']} handleFieldChange={this.handleFieldChange} errorMsg={this.state.errors['title']} />
+          <TextField  objectType="item" name="description" required="" title="Description" value={this.state.formValues['description']} handleFieldChange={this.handleFieldChange} errorMsg={this.state.errors['description']} />
+          <TextField objectType="item" name="transcription" required="" title="Transcription" value={this.state.formValues['transcription']} handleFieldChange={this.handleFieldChange} errorMsg={this.state.errors['transcription']}  />
+          <StringField  objectType="item" name="manuscript_url" required="" title="Digitized Manuscript URL" value={this.state.formValues['manuscript_url']} handleFieldChange={this.handleFieldChange} placeholder="http://" help="Link to externally hosted manuscript viewer." errorMsg={this.state.errors['manuscript_url']}  />
```
jaronkk: Can all of the `objectType="item"` be changed to use a prop that defaults to "item"?  Something like `objectType={this.props.objectType}`

app/assets/javascripts/components/item_meta_data_form/ItemMetaDataForm.js.jsx
```diff
@@ -19,10 +20,34 @@ var ItemMetaDataForm = React.createClass({
...
+    event.preventDefault();
+
+    $.ajax({
+      url: this.props.url,
+      dataType: "json",
+      type: "POST",
+      data: this.postParams(),
+      success: (function(data) {
+        this.setErrors(false);
+        alert("You did it! Item SAVED. ");
```
jaronkk: Can we choose a better confirmation text for this?

app/assets/javascripts/components/Form.js.jsx
```diff
@@ -28,12 +29,24 @@ var Form = React.createClass({
     }
   },
 
+  formErrorAlert: function () {
+    if (this.props.hasErrors)  {
+      return (
+        <div className="alert alert-danger" role="alert">
+          <strong>Opps!!</strong> errors prohibited this form being saved.
```
jaronkk: "Oops" is spelled wrong!  Can we have a better sentence here?  Just some boring generic sentence is fine with me.

app/assets/javascripts/components/item_meta_data_form/ItemMetaDataForm.js.jsx
jonhartzler: there is a later ticket that deals with the un-built messaging system.

jaronkk: Okay, sounds good.

app/assets/javascripts/components/item_meta_data_form/ItemMetaDataForm.js.jsx
jonhartzler: I am not sure.  I intend for these objects to be a something that later can be auto generated by passing a data control structure.  I have been trying to keep the objects to be somewhat generic to help sort out the pattern.

app/assets/javascripts/components/Form.js.jsx
jonhartzler: Do we need this message if we follow 
http://www.google.com/design/spec/patterns/errors.html#errors-user-input-errors?

Also: http://uxmovement.com/forms/how-to-make-your-form-error-messages-more-reassuring/

jgondron: We definitely need to provide the specific errors per field, but I also think we should provide an overall indication that there was some error, especially if all fields are not always visible on one screen. Otherwise you could have a field offscreen that has an error and not notice it. So we may still need both.

jonhartzler:
Consent is that I can merge this branch.

PR #124, opened 2015-05-21
update the form to track if it is dirty or not.

jonhartzler:
We need to message the user when they leave a page with out saving.  This had to be modified from the jquery model we were using to a react model. 

Here I just add a bit of code that adds to the onbeforeunload event when the form is dirty.


app/assets/javascripts/components/item_meta_data_form/ItemMetaDataForm.js.jsx
```diff
@@ -70,6 +71,23 @@ var ItemMetaDataForm = React.createClass({
     })
   },
 
+  setDirty: function () {
+    this.setState({
+      dirty: true,
+    });
+    window.onbeforeunload = function () {
```
jaronkk: This should be done using `window.addEventListener("beforeunload", this.unloadWarning)`, and then removed with window.removeEventListener.  You'll need to define this.unloadWarning (should be the same as the function you already have)

jaronkk:
:+1:

danthewolfe:
:+1:

PR #125, opened 2015-05-21
Dec 349

jonhartzler:
We needed to add a message system so that we can tell users that they have successfully saved the form.  This I think is still a stop gap.  It works with the current form. at of the layout.  I think we might want to look at doing a fixed header similar to gmail or many other web apps.  Then we would wish to change the messaging.  

1. I refactored all the form objects to be in a forms directory. 
2. I added FormSavedMsg and code to show the message when it was saved.  
3. I also implemented the form so that any additional changes remove the message from the display.  It could be timed.  
4. I added an additional ticket to make the messages fade in and out so that the change is not so jerky at a later date.


app/assets/javascripts/components/forms/FieldHelp.js.jsx
```diff
@@ -1,6 +1,6 @@
-//app/assets/javascripts/components/modal/Modal.jsx
+//app/assets/javascripts/components/forms/FieldHelp.jsx
 
-var FormHelpBlock = React.createClass({
+var FieldHelp = React.createClass({
   displayName: 'FormHelpBlock',
```
jaronkk: The displayName is wrong, but is there a reason we need to use the displayName attribute? It defaults to the class name, so I'd vote for just removing it altogether unless we need to override it for some reason.

jonhartzler: I agree.  I will delete.

danthewolfe:
:+1:

jaronkk:
:+1:

PR #126, opened 2015-05-21
Process uploaded images in the background

jaronkk:
Ignore this pull request for now.


rfox2:
:+1: Can we merge this in now? I think that what we're doing with images has evolved a bit since this was done but we're still processing images in the background correct?

PR #127, opened 2015-05-22
update the errors to process the arrays

jonhartzler:
I did not take into account that the errors were arrays.  

Changes to process errors as an array.


jaronkk:
:+1:

PR #129, opened 2015-05-22
simple update to change that the way that panels are built.

jonhartzler:
The way that react panels are generated are not flexible enough. 

I added several classes to handle the different parts of a panel to make it easier use the parts of a panel you wish to display.


jaronkk:
When I load this form I get a number of warnings from react about invalid props.

jaronkk:
:+1:

PR #135, opened 2015-05-27
DEC-328: Further change of title->name

jgondron:
WHY
The name/title discrepancy between the schema and the front end is becoming more apparent as we move more things into the api. For example, V1::ItemsController.update now renders the json with errors when there are errors. The problem is that the item partial calls this field "name", but the errors will return "title" when you send a put with an invalid title. These need to match so that any client applications can properly handle the error without having to do this mapping of title->name
HOW
There are various band-aidy ways we could have done this but we decided it would be best to change it all the way down to the schema for a number of reasons. So the basic process was to change the schema for each entity, then fix what broke. Additionally, I had to change a few services that relied on every object having a title field, namely the creation of urls. Prior to this change CreateBeehiveURL relied on all objects having a 'title' field for creating the slug. In some instances this field was changed to 'name' but in others it was actually changed to 'name_line_1', so it's no longer a consistent field name across all objects. I moved the decision of what to use as the slug down to the model for now and created a ticket https://jira.library.nd.edu/browse/DEC-357 to examine this as a whole.


jonhartzler:
:+1:

jaronkk:
:+1:

PR #136, opened 2015-05-27
Fix showcase saveing params

jonhartzler:
Why 
The subtitle value is not saving on the showcases

How 
Added it to the params listed in the controller.  Then made is show up on the display in the editor.  
this will be in conflict with changes going into phase2.


jgondron:
:+1:

jaronkk:
:+1:

PR #137, opened 2015-05-27
DEC-359: Updating an exhibit does not show changes on refresh

jgondron:
WHY: Problem was with v1::showcases#index. View used exhibit data but cache key generation was not including this. 
HOW: Changed the v1::showcases#index to use the associated exhibit when generating the cache key.


jaronkk:
:+1:

jonhartzler:
:+1:

PR #138, opened 2015-05-28
DEC-262: Allow the copyright text to be changed

jgondron:
WHY: The user may need to include additional copyright details for their exhibit
HOW: Added a copyright field to the exhibits schema and modified honeycomb to allow editing of this field. If the user specifies a custom copyright, it will use it, otherwise, it will use the previously defined copyright text. As an additional note, when the user first creates an exhibit, it will create a default value so that the copyright year is associated with the year in which the exhibit was originally created, as opposed to a dynamically changing value. They can of course always modify this after it's created.

### DIFF
--- a/app/decorators/v1/collection_json_decorator.rb
+++ b/app/decorators/v1/collection_json_decorator.rb
@@ -36,8 +36,9 @@ module V1
     end
 
     def copyright
-      if(!object.exhibit.copyright.to_s.empty?)
-        object.exhibit.copyright.to_s
+      val = object.exhibit.copyright.to_s
+      if(!val.empty?)
+        val
       else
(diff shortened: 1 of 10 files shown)


app/decorators/v1/collection_json_decorator.rb
```diff
@@ -35,6 +35,15 @@ def about
       object.exhibit.about.to_s
     end
 
+    def copyright
+      val = object.exhibit.copyright.to_s
+      if(!val.empty?)
```
jaronkk: Can you fix this hound complaint?

danthewolfe:
:+1:

jaronkk:
:+1:

PR #147, opened 2015-06-03
DEC-355: Deprecation warning

jgondron:
We were receiving: 
DEPRECATION WARNING: `#column_for_attribute` will return a null object for non-existent columns in Rails 5. Use `#has_attribute?` if you need to check for an attribute's existence. (called from _app_views_showcases__form_html_erb__212708199469747219_70292566834860 at /Users/jgondron/repos/honeycomb/app/views/showcases/_form.html.erb:10)

Fixed by updating simple form to 3.1.0 from 3.1.0rc1


jaronkk:
:+1:

danthewolfe:
:+1:

PR #148, opened 2015-06-04
DEC-389: Get showcases dependent on item

jgondron:
Need a way of getting a json from the api for all showcases that use an item. To do this, I added a new route v1/items/:id/showcases which hits ItemsController.showcases. This action will render a json with the following structure:
Collection
--Item
----Showcases
where Showcases will only contain showcases that use that item.


jaronkk:
:+1:

jonhartzler:
:+1:

PR #149, opened 2015-06-04
DEC-394 - Add metadata fields to item

jonhartzler:


app/models/item.rb
```diff
@@ -1,4 +1,6 @@
 class Item < ActiveRecord::Base
+  store :metadata, accessors: [:creator, :publisher, :alternateName, :rights, :originalLanguage]
```
jaronkk: Should we stick with rails convention and use underscores rather than camelcase? Or are we doing it for the javascript side of things?

jonhartzler: Yes we should I copied and pasted out of the api document.

jaronkk: And should we be storing this as json?

jonhartzler: yes but as a note if we go to postgres this will still require a migration. 

NOTE - If you are using PostgreSQL specific columns like hstore or json there is no need for the serialization provided by store. Simply use store_accessor instead to generate the accessor methods. Be aware that these columns use a string keyed hash and do not allow access using a symbol.

jaronkk:
:+1:

db/migrate/20150604150434_add_metadata_to_items.rb
```diff
@@ -0,0 +1,5 @@
+class AddMetadataToItems < ActiveRecord::Migration
+  def change
+    add_column :items, :metadata, :text, limit: 4294967295
```
jgondron: Any reason why we wouldn't do this in another table? Ex: metadata( item_id, key, value)

jonhartzler: simplicity. We get basically the same functionality with out having to build something.  In addition we are looking to migrate to postgres and hstore which works similar to this.

jgondron:
:-1: :+1:

PR #150, opened 2015-06-04
DEC-388: Showcase order caching

jgondron:
WHY: Changing the order of showcases does not show changes on front end. 
HOW:
-Had to start passing the decorated showcase to the cache key generator so that it could pull in the next showcase when generating the cache key.
-Changed the specs now that the cache key generator uses a ShowcaseJSONDecorator object and not just a Showcase object.


jaronkk:
:+1:

jonhartzler:
:+1:

PR #152, opened 2015-06-05
Dec 315 add react fields

jonhartzler:
Update to have a select box to add 5 new meta data fields to item.  

The things I want to say about this branch are more informational since I am leaving.  I think I would like to get it merged in while I am gone so that master does not drift so much.  There are several unpolished parts that I have added tickets for.  
1.  It is a select box and it should probably be a drop button.
2.  The select box does not disappear when there are no items in it. 
3.  new fields appear in the order they are listed in the additionalFieldConfiguration hash and not at the bottom of the form.  
4.  I also think the mechanism for doing this needs some refactoring but I am expecting that to happen with the other tickets revolving around the metadata display.


jgondron:
:+1:

danthewolfe:
:+1:

PR #153, opened 2015-06-10
Dec 264 token auth

rfox2:
**What** Added token authentication for API requests

**Why** To prepare for future functionality that will involve authenticated access to restricted collections

**How** The main components that were added are two classes that either create or validate an auth token. Auth tokens are created for every authenticated user. That token can then be used to validate requests from remote systems to restricted collection content. This could be accomplished several ways: via MQ mechanism, remote invocation via callbacks, etc.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -94,7 +94,6 @@ group :development, :test do
   gem "growl"
   gem "ruby_gntp"
   gem "growl-rspec"
-  gem "growl_notify"
 
   gem "coveralls", require: false
 
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,7 @@ gem "rb-readline"
 
 gem "devise"
 gem "devise_cas_authenticatable"
+gem "activerecord-session_store"
 
 gem "addressable"
 
--- a/Gemfile
+++ b/Gemfile
@@ -73,6 +73,7 @@ gem "autoprefixer-rails"
 group :development, :test do
   gem "rubocop", require: false
 
+  gem "awesome_print"
   gem "i18n-debug"
   gem "pry"
   gem "pry-nav"
@@ -90,9 +91,10 @@ group :development, :test do
   gem "guard-rails"
   gem "guard-rspec"
   gem "guard-spring"
-  #  gem "growl"
-  #  gem "ruby_gntp"
-  #  gem "growl-rspec"
+  gem "growl"
+  gem "ruby_gntp"
+  gem "growl-rspec"
+  gem "growl_notify"
 
   gem "coveralls", require: false
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -164,8 +164,6 @@ GEM
     growl-rspec (3.0.0)
       growl
       rspec (~> 3.0)
-    growl_notify (0.0.3)
-      rb-appscript
     guard (2.6.1)
       formatador (>= 0.2.4)
       listen (~> 2.7)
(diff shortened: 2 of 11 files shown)


jaronkk:
Is this going to change how all session information is stored by moving it to the database rather than the cookie?

rfox2:
It will. My justification for that is that there's an inherent limitation of 4K for cookie data, and I also don't want to store an auth token in a browser cookie that only offers base64 encoding for obfuscation. The session cookie will still contain the session token but nothing else. Is that problematic?

jaronkk:
By design we shouldn't be storing anything more than simple strings in the session data.  The cookie store also does encryption on the cookie, not obfuscation.  The cookie can only be decrypted if you know the secret_key_base key stored in the secrets.yml, so there's no risk to the session information being compromised on the client side.

The cookie store offers performance improvements over other options, since it doesn't have to connect to a database or read files.  See http://api.rubyonrails.org/classes/ActionDispatch/Session/CookieStore.html for more information

rfox2:
I took into consideration the performance issues, which are negligible I think. It's worth the hit for a couple of reasons.

* If the application has ready access to the auth token via the database, then validation can take place. If it's only stored in a cookie, how can back end validation happen? If we set up authenticated access via Beehive in an iframe or via another method, being able to pass the auth token around securely will be required. Either Beehive will need to validate the token somehow, and thus have access to the token value, or Honeycomb will need to validate a request coming from Beehive, and need access to the token value (which would be in the cookie), but how would Beehive get the validation token? Via a GET request from Honeycomb? Yuck.

* That's true about cookie encryption but I'd rather not store an auth token in a cookie. There are ways to decrypt the cookie even in Rails 4, and I'd rather not chance it. What I'm proposing here is a new way to pass secure auth tokens between applications, so it's a different paradigm. If someone was able to get access to a non-published collection, it wouldn't be the end of the world, but I think this model could work for other purposes.

Here are some posts with food for thought about cookies and security:

* http://dev.housetrip.com/2014/01/14/session-store-and-security/
* http://wonko.com/post/why-you-probably-shouldnt-use-cookies-to-store-session-data

If performance is really the driver for us over security, then I'm willing to give up the ghost on this. We're not Amazon, though, so I'm willing to live with a few extra milliseconds per request to have the added security. Plus, for our purposes, storing session data on the server has other advantages. Not trying to take us back to 2007... just being cautious.

jgondron:
It's no longer stateless, so scaling will be trickier. I'm sure that there are other solutions to this, and maybe this application does not need to be very scaleable, but I just wanted to note that as an additional problem other than the performance.

jaronkk:
From what I understand, browser security settings will prevent Honeycomb from accessing the cookie on cross-site requests from Beehive.  This is to prevent cross-site-scripting attacks.  Any AJAX request from Beehive to Honeycomb will be sent without any cookie information.  If that's the case (and there's a chance I'm wrong), then either way I think we'll need to be passing some type of token via GET requests.

We may want to spend some time as a team discussing possible approaches to this token authentication.

I'd also like to have a group face-to-face discussion about the session storage issue.

rfox2:
Jaron: you're right - that's why I suggested that we use an MQ server for this purpose or setting up a callback which would completely back end access to the token.

We can set the domain of the cookie so that it will be sent to *.library.nd.edu as opposed to honeycomb.library.nd.edu, so it would get sent to both systems, but with it being a session cookie Beehive would ignore it anyway unless we wrote a separate cookie with the token in it.

If we use an iframe from within Honeycomb, would it not send the cookie as a part of that request?

jaronkk:
There are about a bajillion ways this can be implemented, so I've scheduled some time tomorrow morning for us to talk about it, since we're responding to multiple points in each comment which probably isn't the most efficient form of communication!

rfox2:
I've thought of another solution to this problem that doesn't involve session data. We can discuss is tomorrow.

rfox2:
I'm going to close this pull request because we're taking a completely different approach to the problem we were trying to solve, but I'll save the code in case we want to take a second look at token auth.

PR #154, opened 2015-06-12
DEC-392: Add showcases to the item edit view

jgondron:
Why: Need to add the list of showcases that are using the item to the item edit view
How:
- Created a ShowcasesPanel to handle rendering the showcase list. This is not tied to items in anyway, so it can be reused to show other brief showcase lists
- Added a title concat mixin (from beehive) to handle generation of the full name of the showcase
- Since we are unable to use the api for this yet, I had to recreate the json generation that was happening inside of the v1 controllers, except this one uses id instead of unique id and links to the non v1 controllers. So ItemDecorator now has a showcases_json method that renders a json with all the needed data for the item's related showcases

### DIFF
--- a/app/assets/javascripts/components/ShowcasesPanel.js.jsx
+++ b/app/assets/javascripts/components/ShowcasesPanel.js.jsx
@@ -1,0 +1,65 @@
+/** @jsx React.DOM */
+
+var ShowcasesPanel = React.createClass({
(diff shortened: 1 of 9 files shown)


jaronkk:
:+1:

rfox2:
:+1:

PR #155, opened 2015-06-12
DEC-337: Remove update from ItemsController 

jgondron:
Why: Item update is now handled in the v1 controller
How: Removed the action from the previous ItemsController and updated the specs to match.


rfox2:
:+1: Stylistically, this is just pristine Justin! The code changes are so elegant.

jaronkk:
:+1:

PR #156, opened 2015-06-12
Dec 371 item model cleanup

rfox2:
**WHAT** Some methods were removed from the Item model because they had inherent redundancies

**WHY** To simplify the Item model

**HOW** I removed the two methods in question, and changed some of the view and decorator code so that the service classes are called directly instead of having callbacks embedded in the Item model

### DIFF
--- a/app/views/v1/items/_item.json.jbuilder
+++ b/app/views/v1/items/_item.json.jbuilder
@@ -3,7 +3,7 @@ json.set! "@type", "CreativeWork"
 json.set! "@id", item_object.at_id
 json.set! "isPartOf/collection", item_object.collection_url
 json.id item_object.unique_id
-json.slug item_object.slug
+json.slug CreateURLSlug.call(item_object.name)
 json.name item_object.name
 json.description item_object.description.to_s
 json.image item_object.image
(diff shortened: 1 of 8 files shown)


app/views/v1/items/_item.json.jbuilder
```diff
@@ -3,7 +3,7 @@ json.set! "@type", "CreativeWork"
 json.set! "@id", item_object.at_id
 json.set! "isPartOf/collection", item_object.collection_url
 json.id item_object.unique_id
-json.slug item_object.slug
+json.slug CreateURLSlug.call(item_object.name)
```
jaronkk: We still want this one to be `item_object.slug` - it's referencing V1::ItemJSONDecorator#slug

danthewolfe:
:+1:

app/views/v1/items/_item.json.jbuilder
rfox2: OK, I'll change that back

jgondron:
Can we make our pattern of usage similar to the other path helpers? ex: 
we can do h.item_edit_path(@item)
so can we do CreateBeehiveURL.item_path(@item)

or even better, if we can extend the built in helper, we could do something like
h.beehive_item_path(@item)
but this might be out of the scope of this ticket.

jaronkk:
@jgondron I think that's out of scope for this ticket

PR #157, opened 2015-06-12
DEC-421: Add strong item delete warning

jgondron:
**Why**: We are adding a list of related showcases to the item edit view, but we also want to additionally give a stronger warning to the user near the delete button if the item is in use by a showcase.

**How**: There are several pages that use the DeletePanel object to render this panel. I extended this class to allow injection of a query object that implements a can_delete method. I wanted to make this as abstract as possible to let those queries answer a high level question of whether or not an object can or cannot be deleted. I also changed it to disable the delete button for now until we implement a cascading delete for items. Once we implement that it will be easy enough to change the warning message in en.yml to give a strong message about what will happen when they hit delete.

In the future, we may need to also change this a bit more to allow those queries to customize the message that gets printed. For instance, if it checks multiple associations, then which one(s) have related records? For now I didn't think this was necessary.

### DIFF
--- a/app/queries/item_query.rb
+++ b/app/queries/item_query.rb
@@ -27,7 +27,7 @@ class ItemQuery
     relation.find_by!(unique_id: id)
   end
 
-  def can_delete
+  def can_delete?
     relation.showcases.count == 0
   end
 end
--- a/app/queries/item_query.rb
+++ b/app/queries/item_query.rb
@@ -26,4 +26,8 @@ class ItemQuery
   def public_find(id)
     relation.find_by!(unique_id: id)
   end
+
+  def can_delete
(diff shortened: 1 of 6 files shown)


app/queries/item_query.rb
```diff
@@ -26,4 +26,8 @@ def published
   def public_find(id)
     relation.find_by!(unique_id: id)
   end
+
+  def can_delete
```
jaronkk: Can you rename this method to `can_delete?` (in general we want any methods that return booleans to have a question mark at the end)

jaronkk:
Other than my comment on the diff, :+1:

rfox2:
:+1: Cool. Will merge.

PR #158, opened 2015-06-12
Process uploaded images in the background.

jaronkk:
### What
Added sneakers (RabbitMQ) as an ActiveJob adapter.  
Images are first uploaded to an uploaded_image paperclip field, then run through initial preprocessing (maximum pixel count, converting tiff to jpg) before being saved locally and then sent to the remote image tile server.
Images being sent to the remote server are limited to one request at a time.
Background processing is disabled by default in development and test.

### Why
Large images take a long time to process.  By uploading the images to a temporary location without doing any processing, the user receives quicker feedback that their action was successful.
The preprocess step in the background allows us to limit the maximum size of images (currently 16000000 pixels), as well as convert TIFFs to a friendlier format.
The remote tile server was having problems processing multiple images simultaneously.  The problem is mitigated by limiting that part of the process to a single worker.

### DIFF
--- a/app/services/preprocess_image.rb
+++ b/app/services/preprocess_image.rb
@@ -1,0 +1,79 @@
+class PreprocessImage
+  MAX_PIXELS = 16000000
+  attr_reader :paperclip_attachment
+
+  def self.call(paperclip_attachment)
+    new(paperclip_attachment).process
+  end
+
+  def initialize(paperclip_attachment)
+    @paperclip_attachment = paperclip_attachment
+  end
+
+  def process
+    if processing_needed?
+      preprocess_attachment
+    end
+    attachment_path
+  end
+
+  private
+
+  def preprocess_attachment
+    processor_attachment.reprocess!
+  end
+
+  def uploaded_image
+    paperclip_attachment
+  end
+
+  def processing_needed?
+    uploaded_image.exists? && (tiff? || exceeds_max_pixels?)
+  end
+
+  def tiff?
+    uploaded_image.content_type == "image/tiff"
+  end
+
+  def exceeds_max_pixels?
+    original_pixels > MAX_PIXELS
+  end
+
+  def original_dimensions
+    @original_dimensions ||= FastImage.size(uploaded_image.path)
+  end
+
+  def original_pixels
+    original_dimensions.inject(:*)
+  end
+
+  def processor_attachment
+    @processor_attachment ||= Paperclip::Attachment.new(:uploaded_image, uploaded_image.instance, processor_options)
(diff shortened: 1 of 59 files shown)


jgondron:
:+1:

jaronkk:
@jonhartzler Can you review this when you have time?

app/assets/javascripts/components/showcase_editor/Item.js.jsx
```diff
@@ -27,6 +27,9 @@ var Item = React.createClass({
 
   render: function() {
     var honeypot_image = this.props.item.links.image;
+    if(honeypot_image == null)
```
jonhartzler: I think we want to have brackets always on control structures.

jaronkk:
@jonhartzler I updated that if statement

app/services/preprocess_image.rb
```diff
@@ -0,0 +1,77 @@
...
+    @paperclip_attachment = paperclip_attachment
+  end
+
+  def process
+    if processing_needed?
+      preprocess_attachment
+    end
+    attachment_path
+  end
+
```
jonhartzler: I would like to see private methods listed as private.

spec/services/save_exhibit_spec.rb
```diff
@@ -4,22 +4,24 @@
   subject { described_class.call(exhibit, params) }
   let(:exhibit) { Exhibit.new }
   let(:params) { { name: "name" } }
-  let(:image) { File.new(Rails.root.join("spec/fixtures/test.jpg").to_s) }
+  let(:upload_image) { Rack::Test::UploadedFile.new(Rails.root.join("spec/fixtures/test.jpg"), "image/jpeg") }
 
   before(:each) do
     allow(exhibit).to receive(:save).and_return(true)
   end
```
jonhartzler: This exhibit is tested well enough for the image part but the spec is incomplete as far as the rest of the saving.

app/workers/retry_worker.rb
```diff
@@ -0,0 +1,32 @@
+require "sneakers/handlers/maxretry"
```
jonhartzler: I am not finding a spec for this file.

app/workers/uploaded_image_worker.rb
```diff
@@ -0,0 +1,8 @@
+class UploadedImageWorker < RetryWorker
```
jonhartzler: I also don't see a spec for this file though I am not sure what you might test.

jonhartzler:
:+1:

PR #159, opened 2015-06-15
Bypassing validation for seed data.

jgondron:
In order to skip the map_user call for seed data

PR #160, opened 2015-06-15
DEC-434 Fix for failed API call preventing logging in to the site

jaronkk:
Fixes a bug where login could be blocked when an exception is encountered connecting to the API server. Also fixes #159


jgondron:
:+1:

rfox2:
:+1:

PR #161, opened 2015-06-15
Cleanup of collection model

rfox2:
**WHAT**

* Remove slug and beehive_url methods to avoid redundancy
* Remove call to slug method
* Remove call to slug method
* Remove call to slug method
* Remove call to beehive_url method

**WHY** This is the same sort of cleanup I did on the item model


jaronkk:
:+1:

jgondron:
:+1:

PR #163, opened 2015-06-16
DEC-327: Auth check v1 items update

jgondron:
**Why:** Permissions check for editing the item was temporarily disabled, needed to be restored.
**How:** For the most part this was the same as any other permissions check, but we determined that there needed to be additional info in the response so that the front end could better report why the user received the response that it did. The ItemMetaDataForm already had mechanisms for rendering errors encountered with item validation. I expanded this to also handle generic server error messages and changed APIController to provide a permission error message when the user is forbidden to make the change.


rfox2:
:+1:

danthewolfe:
:+1:

PR #164, opened 2015-06-16
DEC-425: Add permission check to collection publish/unpublish methods

jgondron:
Why: v1 collections publish/unpublish is currently not enforcing that the user have permission to edit the collection.
How: These actions now bail if the api renders a forbidden response.


rfox2:
:+1:

danthewolfe:
:+1:

PR #165, opened 2015-06-17
Dec 407 collection preview

rfox2:
**WHAT** Added functionality for managing a preview mode for collections. When preview mode is turned on, the collection remains hidden from the DEC home page but can be accessed via a direct URL.

**HOW** Added a service class to manage the preview mode attribute on collections, and added a new boolean field on collections. Added a react component for switching preview mode on and off. Modified the public_find() method for collection queries so that it accounts for the preview mode.

**WHY** We needed a way to enable preview mode so that the preview link on the admin interface would work, and to allow editors to share an accessible link with other individuals who are not editors so that they can view the work in progress prior to it being published.

### DIFF
--- a/app/assets/javascripts/components/CollectionPreviewModeToggle.js.jsx
+++ b/app/assets/javascripts/components/CollectionPreviewModeToggle.js.jsx
@@ -3,7 +3,8 @@
 var CollectionPreviewModeToggle = React.createClass({
   propTypes: {
     collection: React.PropTypes.object.isRequired,
-    previewModePath: React.PropTypes.string.isRequired
+    previewModePath: React.PropTypes.string.isRequired,
+    onToggle: React.PropTypes.func
   },
   getInitialState: function() {
     return {
@@ -14,12 +15,16 @@ var CollectionPreviewModeToggle = React.createClass({
   handleClick: function () {
     this.togglePreviewMode();
   },
+  stateChanged: function() {
+    if(this.props.onToggle)
+      this.props.onToggle(this.state.preview_mode);
+  },
   previewLabel: function (preview_status) {
     var label;
     if (preview_status) {
-      label = 'Preview Enabled'
+      label = 'Enabled'
     } else {
-      label = 'Preview Not Enabled'
+      label = 'Disabled'
     }
     return label;
   },
@@ -41,7 +46,7 @@ var CollectionPreviewModeToggle = React.createClass({
         this.setState({
           preview_mode: preview_state,
           preview_mode_label: this.previewLabel(preview_state)
-        });
+        }, this.stateChanged);
       }).bind(this),
       error: (function(xhr, status, err) {
         console.error(actionUrl, status, err.toString());
--- a/app/assets/javascripts/components/CollectionPreviewModeToggle.js.jsx
+++ b/app/assets/javascripts/components/CollectionPreviewModeToggle.js.jsx
@@ -1,0 +1,59 @@
+/** @jsx React.DOM */
+
+var CollectionPreviewModeToggle = React.createClass({
+  propTypes: {
+    collection: React.PropTypes.object.isRequired,
+    previewModePath: React.PropTypes.string.isRequired
+  },
+  getInitialState: function() {
+    return {
+      preview_mode: this.props.collection.preview_mode,
+      preview_mode_label: this.previewLabel(this.props.collection.preview_mode)
+    };
+  },
+  handleClick: function () {
+    this.togglePreviewMode();
+  },
+  previewLabel: function (preview_status) {
+    var label;
+    if (preview_status) {
+      label = 'Preview Enabled'
+    } else {
+      label = 'Preview Not Enabled'
+    }
+    return label;
+  },
+  togglePreviewMode: function () {
+    var preview_state;
+    if (this.state.preview_mode) {
+      preview_state = false;
+    } else {
+      preview_state = true;
+    }
+    $.ajax({
+      url: this.props.previewModePath,
+      dataType: "json",
+      data: {
+        value: preview_state
+      },
+      method: "PUT",
+      success: (function(data) {
+        this.setState({
+          preview_mode: preview_state,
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/components/CollectionPreviewModeToggle.js.jsx
```diff
@@ -0,0 +1,59 @@
...
+      error: (function(xhr, status, err) {
+        console.error(actionUrl, status, err.toString());
+      }).bind(this)
+    });
+},
+render: function () {
+  return (
+    <label>
+    <input type="checkbox" checked={this.state.preview_mode} onChange={this.handleClick}/>
+    <span className="toggle"></span><span className="toggle-label">{this.state.preview_mode_label}</span>
```
jaronkk: Rather than track the label via the state, can you change it so the previewLabel function returns a label based on `this.state.preview_mode`, rather than accepting preview_status as an argument?  Currently `this.state.preview_mode_label` is redundant since you can always calculate it based on the current state.

app/assets/javascripts/components/CollectionPreviewModeToggle.js.jsx
```diff
@@ -0,0 +1,59 @@
+/** @jsx React.DOM */
+
+var CollectionPreviewModeToggle = React.createClass({
+  propTypes: {
+    collection: React.PropTypes.object.isRequired,
+    previewModePath: React.PropTypes.string.isRequired
+  },
```
jaronkk: Can you add an additional line break here, and after each function in this component?  We haven't really formalized that as a standard, but I've found it more readable.

Example: https://github.com/ndlib/honeycomb/blob/master/app/assets/javascripts/components/Modal.js.jsx

app/queries/collection_query.rb
```diff
@@ -22,7 +22,10 @@ def public_collections
   end
 
   def public_find(id)
-    relation.find_by!(unique_id: id, published: true)
+    relation.where(
+      "unique_id = ? AND (published = ? OR preview_mode = ?)",
+      id, true, true
+    ).take!
```
jaronkk: I feel like I should have known about the take method before... Nice!

app/services/set_collection_preview_mode.rb
```diff
@@ -0,0 +1,18 @@
...
+  def initialize(collection, value)
+    @collection = collection 
+    @value = value
+  end
+
+  def set_preview_mode
+    collection.preview_mode = value
+    collection.save
+  end
+end
```
jaronkk: Need a newline at the end of the file.

spec/services/set_collection_preview_mode_spec.rb
```diff
@@ -0,0 +1,30 @@
+require "rails_helper"
+
+describe "SetCollectionPreviewMode" do
+  let(:collection) { double(Collection, "preview_mode=" => false, save: true) }
```
jaronkk: In general we should use instance_double instead of double wherever we can.

spec/services/set_collection_preview_mode_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      subject.set_preview_mode
+    end
+
+    it "should save the collection record" do
+      expect(collection).to receive(:save).and_return(true)
+      subject.set_preview_mode
+    end
+
+    it "should return false if the save fails" do
+      allow(collection).to receive("save").and_return(false)
```
jaronkk: Change this to `receive(:save)` for consistency's sake.

jaronkk:
I added a lot of comments to specific lines, but really this is great.  Once my comments and any remaining hound violations are addressed, :+1:

rfox2:
Justin is going to get these changes in and then add a bit of functionality to hide or show the Preview Site link. Thanks Jaron!

app/assets/javascripts/components/CollectionPreviewModeToggle.js.jsx
```diff
@@ -0,0 +1,62 @@
+/** @jsx React.DOM */
+
+var CollectionPreviewModeToggle = React.createClass({
+  propTypes: {
+    collection: React.PropTypes.object.isRequired,
+    previewModePath: React.PropTypes.string.isRequired,
+    onToggle: React.PropTypes.func
+  },
```
jaronkk: We need line breaks between functions in this component as well

jgondron: Fixed. Also fixed a spacing issue with the render function

jaronkk:
:+1:

app/assets/javascripts/components/CollectionPreviewPublish.js.jsx
```diff
@@ -0,0 +1,61 @@
...
+  getInitialState: function() {
+    return {
+      published: this.props.collection.published,
+      preview: this.props.collection.preview_mode,
+    };
+  },
+
+  linkLabel: function () {
+    if (this.state.published)
+      return this.props.liveLinkLabel;
```
jonhartzler: I think this if statement should include { }s

rfox2:
:+1: Definitely! That Rob Fox guy is a genius.

app/assets/javascripts/components/CollectionPreviewPublish.js.jsx
```diff
@@ -0,0 +1,61 @@
...
+    return this.props.previewLinkLabel;
+  },
+
+  siteLink: function() {
+    if(this.state.published || this.state.preview)
```
jonhartzler: Also here { } s and in the else below.

jgondron: Fixed

app/assets/javascripts/components/CollectionPreviewPublish.js.jsx
jgondron: Fixed

jonhartzler:
:+1:

PR #166, opened 2015-06-19
Improve deploy speed, upgrade to Capistrano 3

jaronkk:
Currently the asset precompilation takes over 3 minutes and runs every time a deploy is done.  Rails 4 makes use of tmp/cache when doing the assets:precompile task so it doesn't have to recompile assets if they haven't changed.

This pull request upgrades to Capistrano 3, and adds a symlink for tmp/cache to improve deploy speed.

Also as a bonus fixed the favicon link tag!


danthewolfe:
:+1:

jonhartzler:
:+1:

jgondron:
:+1:

PR #167, opened 2015-06-22
Dec 396 clarify add item modal

jonhartzler:
Why:  Needed to update the messaging in the item creation modal to make it easier to use.  

What: 
1. Reorganized the modal and dropzone interactions so that I can update the titling of the bottom close button.  see ReactDropzone.jsx line 34 and 35
2. Changed the way that the call to action message is displayed in the modal to try and make it clearer about what is happening. 
3. Changed the modal to continue to display the call to action so that the users continue to know what to do. 
4.  Fixed the problem where you could only click once on the modal to add more items from the file system.


jgondron:
:+1:

jaronkk:
:+1:

PR #168, opened 2015-06-22
DEC-446 Add Browserify to project

danthewolfe:
Adding in Browserify to make it easier to make it easier to include external libraries.

Followed adaptations from https://reactjsnews.com/setting-up-rails-for-react-and-jest/ and Beehive's setup.


.hound.yml
```diff
@@ -3,4 +3,5 @@ ruby:
 
 javascript:
   config_file: config/style_guides/.javascript-style.json
+  ignore_file: app/assets/javascripts/components.js
```
jaronkk: I think this is expecting to find a file that specifies a list of javascript files to ignore.  Beehive example: https://github.com/ndlib/beehive/blob/master/config/style_guides/.javascript-ignore

.travis.yml
```diff
@@ -1,10 +1,16 @@
 language: ruby
-cache: bundler
 sudo: false
 rvm:
   - 2.1.4
+install:
+  - bundle install
```
jaronkk: We don't want to specify bundle install, since I think it overrides the way Travis does bundle caching between builds.

jaronkk:
:+1:

jaronkk:
The deployment depends on node being installed on the servers, but I think we can go ahead and merge this. :+1:

jonhartzler:
:+1:

PR #169, opened 2015-06-23
Advanced seed data using external repo

jaronkk:
This creates some seed data for development purposes.

I've tied the classes in to the `rake db:seed` task, but it could be made its own task if we wanted.

It pulls seed data from an external repo at https://github.com/ndlib/honeycomb-brood (so we don't have to store large files in this repo) and creates collections and items as defined in the repo.  It abuses the unique_id field to maintain its relationships, so urls on the front end wouldn't change after dropping/recreating the database.

It could also be updated *relatively* easily to add showcases/sections.

### DIFF
--- a/app/services/brood/collection.rb
+++ b/app/services/brood/collection.rb
@@ -1,0 +1,98 @@
+module Brood
+  class Collection
+    attr_reader :base_path, :item_count
+
+    def initialize(base_path)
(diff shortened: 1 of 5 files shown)


danthewolfe:
:+1:

jonhartzler:
:+1:

PR #171, opened 2015-06-26
DEC-443 display errors

danthewolfe:
* Created a mediator for sending messages to subscribers. Subscribers provide a specific call back function that will be called when a message that they are concerned with is received.
 * Cleaned up the different form messages since they were duplicating code for the most part, aside from the message and style
 * Created a FormMessageCenter that handles displaying all "global" messages to the user. This object subscribes to certain messages for triggering the display of a message, focusing on the message, and hiding it.
 * Created an APIResponseMixin to handle parsing the response json when errors are received from the api to get an easier to use error message
 * Stripped out the error/success messaging parts of ItemMetaDataForm and changed it to utilize the message center object for this
 * Modified CollectionPublishToggle to display an error message when it is unable to publish/unpublish the collection.

======

 * Added and implemented material-ui for the error dialog.
 * Created some mixins for general material-ui theme related stuff and for dialog components.

### DIFF
--- a/app/assets/javascripts/components/forms/FormMessageCenter.jsx
+++ b/app/assets/javascripts/components/forms/FormMessageCenter.jsx
@@ -1,0 +1,58 @@
+var FormMessageCenter = React.createClass({
+  getInitialState: function() {
+    return {
+      messageType: "",
+      messageText: "",
+      displayState: "hidden",
+    };
+  },
+
+  componentWillMount: function() {
+    mediator.subscribe("MessageCenterDisplay", this.receiveDisplay);
+    mediator.subscribe("MessageCenterHide", this.receiveHide);
+    mediator.subscribe("MessageCenterFocus", this.receiveFocus);
+    mediator.subscribe("MessageCenterDisplayAndFocus", this.receiveDisplayAndFocus);
+  },
+
+  receiveDisplay: function(type, message) {
+    this.setState({
+      displayState: "show",
+      messageType: message[0],
+      messageText: message[1],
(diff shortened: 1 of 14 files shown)


app/assets/javascripts/mediator.js
```diff
@@ -0,0 +1,24 @@
+function Mediator() {
+  // hashmap of message type to list of subscribers for that message type
+  this.subscribers = [];
+};
+
+Mediator.prototype.subscribe = function(type, subscriberCallback) {
+  if (!(type in this.subscribers))
+    this.subscribers[type] = [];
```
jaronkk: This if statement needs curly braces

app/assets/javascripts/components/forms/FormMessageCenter.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+  receiveDisplay: function(type, message) {
+    this.setState({
+      messageType: message[0],
+      messageText: message[1],
+    });
+    this.refs.errorDialog.show();
+  },
+
+  receiveDisplayAndFocus: function(type, message) {
```
jgondron: This is no longer needed now that we're using the dialogs as a way to ensure the user sees the message.

rfox2:
:+1: Would the mediator code need to be deprecated if we switch to flux?

danthewolfe:
Probably. I think we'd remove the mediator.js file and use flux's dispatcher instead. It would still be a very similar pattern of subscribing and event emitter.

PR #172, opened 2015-06-26
Fix the roboto font

jaronkk:
The Roboto font is currently referenced via HTTP so it doesn't work when the environment is under HTTPS.


danthewolfe:
:+1:

rfox2:
:+1:

PR #174, opened 2015-07-20
DEC-432: On destroy, cascade

jgondron:
Why: Need to ensure referential integrity when deleting data. We decided we should cascade destroy most objects. The only exception is when a user directly deletes an item, it should constrain on section and child items.
How: Modified the item forms to only allow the user to delete an item if it has no associated sections or child items. Created Destroy service objects that handle cascading the destroy down through all associated objects and changed the controllers over to use these new services. Added papertrail to all models. As a final safeguard created foreign key constraints at the database layer to prevent any mishaps with application logic.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -106,6 +106,8 @@ group :development, :test do
   gem "web-console", "~> 2.0"
 
   gem "factory_girl_rails"
+
+  gem "rails-erd"
 end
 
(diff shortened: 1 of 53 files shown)


jaronkk:
:+1:

jonhartzler:
:+1:

PR #175, opened 2015-07-21
DEC 398

danthewolfe:
What: Replace checkboxes with material-ui toggles.
Why: Using standard components simplifies the code vs. continuously recreating them.

### DIFF
--- a/app/assets/javascripts/components/publish/PublishToggle.jsx
+++ b/app/assets/javascripts/components/publish/PublishToggle.jsx
@@ -1,19 +1,35 @@
 /** @jsx React.DOM */
 var React = require('react');
+var mui = require("material-ui");
+var Toggle = mui.Toggle;
 var PublishToggle = React.createClass({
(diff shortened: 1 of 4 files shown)


jonhartzler:
:+1:

jgondron:
:+1:

app/assets/javascripts/components/publish/PublishToggle.jsx
```diff
@@ -1,19 +1,35 @@
...
   propTypes: {
     publishPanelFieldName: React.PropTypes.string.isRequired,
     published: React.PropTypes.bool.isRequired,
     togglePublished: React.PropTypes.func.isRequired,
   },
+
+  publishLabel: function () {
+    var label;
+    if (this.props.published) {
+      label = 'Publishedxxx'
```
jaronkk: I assume the "xxx" was a test value that we should probably remove.

jaronkk:
After my comment, :+1:

PR #178, opened 2015-07-21
Hiding broken sub resources on item edit page.

danthewolfe:
Description:
I think we should just hide this whole section but that should be raised before it is done.

https://jira.library.nd.edu/browse/DEC-505


jonhartzler:
:+1:

jaronkk:
:+1:

PR #179, opened 2015-07-23
Immediate date fix

jonhartzler:
Why:  The code for validating dates and sending api response was both using the metadata_date model.  That model was incompatible with both operations in the case when data in the form was invalid.  This was because the we were relying on the failure of the metadata date to convert itself to a ruby date to raise exceptions.  This set of changes refactors that.  I also determined that the old storage format for dates caused a lot of extra code to be written so I refactored that as.

What:  I updated the way that date data is stored see: https://github.com/ndlib/honeycomb/wiki/Meta-Data-Storage  and how the date object will be sent in the api see: https://github.com/ndlib/honeycomb/wiki/Meta-Data-Api.   I also adjusted where the validations are happening and to stop relying on the metadata date conversion.  In addition I refactored the code so that the conversion algorithms got their own classes.

### DIFF
--- a/app/values/metadata_date.rb
+++ b/app/values/metadata_date.rb
@@ -7,7 +7,6 @@ class MetadataDate
   validates :month, numericality: { only_integer: true, allow_nil: true, greater_than_or_equal_to: 1, less_than_or_equal_to: 12 }
   validates :day, numericality: { only_integer: true, allow_nil: true, greater_than_or_equal_to: 1, less_than_or_equal_to: 31  }
 
-
   class ParseError < Exception
   end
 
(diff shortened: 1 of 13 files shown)


jaronkk:
:+1:

app/values/metadata_date.rb
```diff
@@ -90,16 +78,83 @@ def date_format
     end
 
     def format_date
-      date = I18n.localize(metadata_date.date, format: date_format)
-      date = fix_bc_date(date)
+      if !date = metadata_date.to_date
+        return ""
+      end
+
+      date = I18n.localize(date, format: date_format)
```
jgondron: I know we have a separate ticket on this, but maybe we should set the timezone in the configs now since we are starting to do more with date parsing.

jgondron:
:+1:

PR #180, opened 2015-07-23
Add html fields

jonhartzler:
Why: we need redactor based fields.

What: I added an html type to the form set.


jaronkk:
:+1:

jonhartzler:
:+1:

PR #184, opened 2015-07-27
DEC-459: Flux

jgondron:
**Why:** We need a better way for components to communicate on the front end. We currently have multiple components that are driven by shared collection properties. Ex: Preview/publish link and the toggles.
**How:** Created a CollectionStore that contains properties title, published, and preview. Components can subscribe to the store to get notification of when it has changed and update it's state accordingly. While components can directly read the properties of the store, they should always use CollectionActions to change both the store and the data on the back end. Additionally, there should no longer be any need for a child component to communicate with a parent component, ex in an onclick type event. As an example of this, I pulled the CollectionPreviewPublish out into it's own separate component. Prior to having a store, this component needed to be the parent of the two toggles in order to stay updated on the state of the collection. Now that we have the store, it can subscribe to listen for changes to the store and update it's link accordingly, so they no longer need to be coupled.


jaronkk:
:+1:

jonhartzler:
:+1:

PR #186, opened 2015-07-28
Guard npm

danthewolfe:
* Adding `npm install` to guardfile with guard-npm.
 * Remember to run `bundle update hesburgh_infrastructure`


jonhartzler:
:+1:

PR #187, opened 2015-07-28
Dec 481 update api

jonhartzler:
Updates to have the api produce data in the new prescribed api format.
https://github.com/ndlib/honeycomb/wiki/Meta-Data-Api


jgondron:
:+1:

app/decorators/v1/metadata_json.rb
```diff
@@ -6,63 +6,63 @@ def self.metadata(item)
...
+    end
 
-      if value.present?
-        value
-      end
+    def date_value(value, label)
+      MetadataDate.new(value.symbolize_keys).to_hash(label)
     end
   end
 end
```
rfox2: Are we still using jbuilder for JSON views? I thought we were backing away from that in favor of hash constructs and to_json.

rfox2:
Just posted a question about jbuilder. Are we still using it and writing new views with it?

app/decorators/v1/metadata_json.rb
jonhartzler: I think the answer is no we are not.  It has never solved problems where there are complex chains very well and these are pretty complex chains.

jaronkk:
:+1:

PR #189, opened 2015-07-31
DEC-511: Google test

jgondron:
I'd like for you guys to pull this and get a feel for how it works before I go any further. I don't intend to merge this just yet, since I still need to add tests and handle specific errors, but I want everyone to understand the flow of this and sign off on it before I spend much more time on it.

The first thing you'll have to do is edit the secrets.yml and populate the client id/secret with the codes from the google dev portal. Once that's done, you should be able to go to collections/:id/import_google_sheet, copy/paste a url to a valid worksheet in your google docs, then click the button. If successful, it should print out the data in the sheet.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -82,6 +82,8 @@ gem "bootstrap-sass", "~> 3.3.1"
 gem "bootstrap-material-design"
 gem "autoprefixer-rails"
 
+gem "google_drive"
+
 group :development, :test do
   gem "rubocop", require: false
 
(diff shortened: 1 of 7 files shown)


jaronkk:
I'm okay with this in general.  Is there a way we can configure access so we're not requesting write permission to the entire Drive account?

Also, if we're going the route of Drive integration, then I think we need to implement their Picker API to give users a browse button: http://googleappsdeveloper.blogspot.com/2012/08/allowing-user-to-select-google-drive.html

jgondron:
@jaronkk yes to both, I've added some of these details to the import ticket. Closing this PR since we all seem to be ok with moving forward on this.

PR #192, opened 2015-08-03
DEC-521: FK constraint checks

jgondron:
Why: Current migration allowed for adding some foreign keys before breaking on others. This lead to a partially migrated state that was difficult to recover from.
How: Expanded the migration to first do checks for any records that would break the migration. If any fail, it will print the broken association and bail out of this and all future migrations. This should help us to isolate which tables we need to check whenever we deploy.


jonhartzler:
:+1:

jaronkk:
:+1:

PR #194, opened 2015-08-04
DEC-507 Index items

jaronkk:
This PR includes code to index Item API data into Solr using Sunspot.

I added a Metadata::Configuration class to contain the configuration information for item metadata.

The Waggle namespace includes all of the code for indexing item data.

Waggle::Item is a wrapper around the item API JSON output.  
Waggle::Metadata::Set is a wrapper around the item's metadata JSON.  It is used to return all metadata field values in their indexable form.
Waggle::Index::Item is responsible for setting up the indexing of Waggle::Item using Sunspot.

This does not include the functionality to index items automatically on save.  That will come in a separate pull request.

### DIFF
--- a/app/models/waggle/index/item.rb
+++ b/app/models/waggle/index/item.rb
@@ -1,0 +1,26 @@
+module Waggle
+  module Index
+    module Item
+      def self.setup(configuration = default_configuration)
+        reset!
+        ::Sunspot.setup(Waggle::Item) do
+          configuration.fields.each do |field|
+            text field.name
+          end
+          string :type, stored: true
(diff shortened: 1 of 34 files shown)


app/models/waggle/index/item.rb
```diff
@@ -0,0 +1,32 @@
+module Waggle
+  module Index
+    module Item
+      def self.setup(configuration = default_configuration)
```
jaronkk: With the way sunspot is configured, there's no easy way to simplify this block.

jonhartzler:
:+1:

PR #195, opened 2015-08-04
DEC-521: Rewrote using direct execution of queries instead of models …

jgondron:
…so that Jaron can sleep at night.


jaronkk:
:+1: :sleeping:

jonhartzler:
:+1:

PR #197, opened 2015-08-04
Add snackbar for page messages

jonhartzler:
I changed the messaging so it always uses the snackbar,  This adds

### DIFF
--- a/app/decorators/display_flash_message.rb
+++ b/app/decorators/display_flash_message.rb
@@ -27,7 +27,7 @@ class DisplayFlashMessage < Draper::Decorator
 
   def display_flash_message(key)
     if object[key].present?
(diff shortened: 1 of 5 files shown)


jaronkk:
:+1:

app/decorators/display_flash_message.rb
```diff
@@ -27,7 +27,7 @@ def flash_keys
 
   def display_flash_message(key)
     if object[key].present?
-      h.render "shared/display_flash_message", content: object[key], css_class: KEY_2_CSS_CLASS[key]
+      h.react_component("PageMessage", messageText: object[key], css_class: KEY_2_CSS_CLASS[key])
```
jgondron: Is the css_class argument needed? Shouldn't there be a messageType instead?

jgondron:
:+1:

PR #199, opened 2015-08-04
Updates from annex

jaronkk:
I was going to write some code that would make use of the NotifyError service, but wanted to get the updated version from the Annex.

I also pulled in the updates to the Sneakers rake task that is being used on Annex, and added the cron task to make sure Sneakers is running.

In MapUserToAPI updated the NotifyError call, and broke down the map! method to satisfy the linter.

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -10,6 +10,9 @@ require "capistrano/rails"
 # Include npm extensions
 require "capistrano/npm"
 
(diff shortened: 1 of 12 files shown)


jonhartzler:
:+1:

jgondron:
:+1:

PR #200, opened 2015-08-05
Add missing beehive title

jonhartzler:
We don't have a way to control if we should show the site title as text in beehive this creates a way.

### DIFF
--- a/app/decorators/v1/collection_json_decorator.rb
+++ b/app/decorators/v1/collection_json_decorator.rb
@@ -44,6 +44,10 @@ module V1
       end
     end
 
+    def show_page_title
+      object.exhibit.show_page_title ? true : false
+    end
+
     def slug
       CreateURLSlug.call(object.name_line_1)
     end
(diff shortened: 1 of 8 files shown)


app/decorators/v1/collection_json_decorator.rb
```diff
@@ -44,6 +44,10 @@ def copyright
       end
     end
 
+    def show_page_title
+      object.exhibit.show_page_title ? true : false
```
jaronkk: Since it's a boolean, it should be `show_page_title?`.  Also, if you use `object.exhibit.show_page_title?` you won't need to use the ternary operator.

spec/decorators/v1/collection_json_decorator_spec.rb
```diff
@@ -43,6 +43,25 @@
     end
   end
 
+  describe "#show_page_title" do
+    let(:exhibit) { double(Exhibit, show_page_title: nil) }
```
jaronkk: We should use instance_double wherever possible

jgondron:
:+1: after Jaron's comments are addressed

jaronkk:
:+1:

PR #201, opened 2015-08-06
DEC-417 Restyle drop down as material-ui component

rfox2:
What: Changed the metadata form menu from a standard select box to a material-ui component; Updated the styles; Refactored the ItemMetaDataForm react component.

### DIFF
--- a/app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
+++ b/app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
@@ -9,9 +9,7 @@ var HtmlField = require('./HtmlField');
 var TextField = require('./TextField');
 var MultipleField = require('./MultipleField');
 
-var Colors = require("material-ui/src/styles/colors");
 var Spacing = require("material-ui/src/styles/spacing");
-var ColorManipulator = require("material-ui/src/utils/color-manipulator");
 
 var fieldTypeMap = {
   string: StringField,
--- a/app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
+++ b/app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
@@ -218,15 +218,8 @@ var ItemMetaDataForm = React.createClass({
   },
 
   render: function () {
-    var dropDownStyle = {
-      height: 40,
-    };
-    var dropDownLabelStyle = {
-      lineHeight: "36px",
-    };
     var dropDownIconStyle = {
-      top: 5,
-      right: Spacing.desktopGutterLess,
+      right: this.muiTheme.spacing.desktopGutterLess,
     };
     var underlineStyle = {
       borderTop: "solid 2px rgb(44, 88, 130)",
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
```diff
@@ -1,5 +1,10 @@
 //app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
 var React = require('react');
+var mui = require("material-ui");
+var Dialog = mui.Dialog;
+var RaisedButton = mui.RaisedButton;
+var DropDownMenu = mui.DropDownMenu;
+var FontIcon = mui.FontIcon;
```
jaronkk: From what I can tell, Dialog, RaisedButton, and FontIcon are not used here.

app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
```diff
@@ -7,6 +12,10 @@ var HtmlField = require('./HtmlField');
 var TextField = require('./TextField');
 var MultipleField = require('./MultipleField');
 
+var Colors = require("material-ui/src/styles/colors");
+var Spacing = require("material-ui/src/styles/spacing");
+var ColorManipulator = require("material-ui/src/utils/color-manipulator");
```
jaronkk: It looks like Colors and ColorManipulator are not being used.  Also see my pull request regarding Spacing here: https://github.com/ndlib/honeycomb/pull/203/files

app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
rfox2: And so uh, you don't like that???

jaronkk:
The other issue I ran into is there's an area at the top of the button that's not clickable for me.  It seems to be related to this padding:

pig-in-mud_-_pretty_animals_-_honeycomb

app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
rfox2: I had been using Colors but removed the references. I can take it out, but I think it's a pain that we need to explicitly add it when we want to reference the theme colors.

rfox2: Not sure I follow on what you did with PR 203. Seems like 6 of one and 1/2 dozen of the other.

rfox2:
Jaron: Something odd about your rendering. The transcription edit field is not present. Could that be the issue?

jaronkk:
@rfox2 With PR 203 I'm accessing values from the HoneycombTheme, rather than default values from material-ui.

`var Spacing = require("material-ui/src/styles/spacing");` will give you the default values, but accessing the spacing through the current theme will give the correct values if they end up being customized.  At least that's how I understand it.  @danthewolfe ?

jaronkk:
The transcription field is up higher.  I have several fields added via the dropdown.

rfox2:
The hesburgh theme currently just references the same thing.... but I'm happy to switch it. ;-)

rfox2:
I fixed the padding issue with the drop down. It's very hard to resize the components due to all of the interaction and potentially conflicting styling. Not worth wasting time on.

app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
```diff
@@ -7,6 +9,8 @@ var HtmlField = require('./HtmlField');
 var TextField = require('./TextField');
 var MultipleField = require('./MultipleField');
 
+var Spacing = require("material-ui/src/styles/spacing");
```
jaronkk: You can eliminate this line now since you're referencing the theme object for the spacing :)

jaronkk:
After that last line is removed, :+1:

jonhartzler:
:+1:

PR #202, opened 2015-08-06
DEC-506 automatically index items on save and destroy

jaronkk:
This adds automatic indexing and removal of items on save and destroy.

I changed v1/item_json_decorator so I can get JSON directly from the model instead of having to make a network request.

I moved all of the Sunspot code into an adapter module under Waggle, and implemented top level methods on Waggle that pass through to the adapter.

I implemented the index service class so it will notify index errors via Airbrake and won't disrupt the save/destroy flow.  This means indexing won't happen in development unless Solr is running.

Solr is not set to auto-start at this point, but I will add that when I connect the search functionality.


app/models/waggle.rb
```diff
@@ -0,0 +1,29 @@
+module Waggle
+  def self.setup
+    adapter.setup
+  end
+
+  def self.adapter
+    self::Adapters::Sunspot
+  end
+
+  def self.index(*objects)
```
rfox2: Given our discussion today, it sounds like the interface will consume specifically formatted JSON instead of objects. I'd recommend changing the terminology so that it makes sense when we need to refactor. I think the long term intent is to make this completely decoupled from the parent application, right? How about:

```ruby
def self.index(*json)
   adapter.index(*json)
end
```

And, just keep that terminology consistent both in Waggle and the adapters we write for it (or, modify the terminology in the adapters to fit the implementation of the chosen indexer).

app/models/waggle.rb
```diff
@@ -0,0 +1,29 @@
...
+    adapter.index(*objects)
+  end
+
+  def self.index!(*objects)
```
rfox2: I understand using the bang methods that correspond to the bang methods in the Sunspot API, but do we want the Waggle interface to correspond to the underlying adapter so closely? Should we parameterize options such as autocommit, or just go with the convention of bang methods indicating autocommit? I was wondering if this locks us in, semantically, to a certain type of implementation. In other words, could we just go with index(json: json, autocommit: true) as opposed to having two methods including index!()?

app/models/waggle/adapters/sunspot.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      def self.remove!(*objects)
+        ::Sunspot.remove!(*objects)
+      end
+
+      def self.commit
+        ::Sunspot.commit
+      end
+
+      def self.register_sunspot_adapters
```
rfox2: Does this need to be exposed as a public method?

app/models/waggle/adapters/sunspot.rb
```diff
@@ -0,0 +1,48 @@
...
+          self::Adapters::InstanceAdapter,
+          Waggle::Item
+        )
+        ::Sunspot::Adapters::DataAccessor.register(
+          self::Adapters::ItemDataAccessor,
+          Waggle::Item
+        )
+      end
+
+      def self.setup_indexers
```
rfox2: Does this need to be a public method?

app/models/waggle/adapters/sunspot.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+      def self.commit
+        ::Sunspot.commit
+      end
+
+      def self.register_sunspot_adapters
+        ::Sunspot::Adapters::InstanceAdapter.register(
+          self::Adapters::InstanceAdapter,
+          Waggle::Item
```
rfox2: Item here is an item in the very generic sense right? This doesn't correspond in any way to DEC items does it?

app/models/waggle/adapters/sunspot/index/item.rb
```diff
@@ -0,0 +1,39 @@
...
+              configuration.fields.each do |field|
+                if field.type == :date
+                  time field.name, multiple: true
+                elsif [:string, :html].include?(field.type)
+                  text field.name
+                else
+                  raise "unknown type #{field.type}"
+                end
+              end
+              string :collection_id, stored: true
```
rfox2: This really weds Waggle to DEC, and a specific implementation of DEC. Do we want that? If not, I think it would be better to parameterize this field list somehow.

app/services/index/item.rb
```diff
@@ -0,0 +1,30 @@
+module Index
```
rfox2: Just curious - should we set up a backgrounder for this (via the message queue)? If so, should any of that be built into this class or be completely separate? I'm not holding the IMS up as a model by any stretch...

jaronkk: I have a separate ticket (DEC-508) to move this process into the background.  But yes, we want this to happen in the background somehow.

app/models/waggle/adapters/sunspot.rb
jaronkk: Right now Waggle::Item corresponds 1:1 to the Honeycomb Item class

app/models/waggle/adapters/sunspot.rb
jaronkk: I've made this a private class method

app/models/waggle/adapters/sunspot.rb
jaronkk: I've made this a private class method

app/models/waggle.rb
jaronkk: It's not a strong preference, but I do like having the bang methods to indicate autocommit vs not.  If we didn't use the bang methods, my preference would be to require the calling code to call commit immediately afterward, rather than adding an autocommit optional argument.

app/models/waggle.rb
jaronkk: Unfortunately it's really not possible to do this type of implementation using Sunspot, so I'd rather add a new issue that handles replacing the Sunspot adapter with a more generic Solr adapter that makes use of RSolr.

PR #203, opened 2015-08-06
Make the current muiTheme available to the component including the mixin

jaronkk:
I think we should be able to get the Spacing object from the current theme using something similar to this structure.


rfox2:
:+1:

PR #206, opened 2015-08-07
DEC-540: Item showcase list caching bug

jgondron:
Why: Item edit page is not updating the list of showcases if a user has previously visited the page
How: Added showcase list to cache key generation for item edit


danthewolfe:
:+1:

jonhartzler:
:+1:

PR #209, opened 2015-08-10
fix the validations on the date field

jonhartzler:
It was requiring month and day.  To fix this I allowed blank in the validations and I then needed to remove  blank rows from the ruby convert to date class.


jaronkk:
Do you need to do the same thing for the convert to iso date?

jaronkk:
:+1:

danthewolfe:
:+1:

PR #211, opened 2015-08-11
fix date bc checked.

jonhartzler:
the bc check box was coming back always checked because the json boolean value is sent as "false" when it is passed through react_component.


app/assets/javascripts/components/forms/DateField.jsx
```diff
@@ -114,7 +113,12 @@ var DateField = React.createClass({
     return "";
   },
 
+  bcChecked: function () {
+    return (this.state.bc === true || this.state.bc === 'true');
+  },
+
   render: function () {
+    console.log(this.state);
```
jgondron: :warning:

jgondron:
:+1: once the log is removed

jaronkk:
:+1: Once the log is removed!

PR #215, opened 2015-08-12
Showcase list bug

jgondron:
Why: If an item belongs to a showcase that does not have a background image, it will fail to render the showcase list in the edit page for that item
How: Changed to use material ui avatars. Will use an avatar image if the showcase has a background image, and an avatar with the first letter of the showcase if it does not.


jaronkk:
:+1:

jonhartzler:
:+1:

PR #217, opened 2015-08-12
DEC-552: Flash messages are improperly cached

jgondron:
Why: Pages that contained flash messages were constantly flashing the message on refresh due to caching.
How: Refactored to use the cacheable_flash gem. This gem stores the flash message in a cookie, and when read, consumes the cookie so that it doesn't get displayed on subsequent reloads.


jaronkk:
:+1:

jonhartzler:
:+1:

PR #218, opened 2015-08-12
Section does not show the correct image

jgondron:
Why: If a section is created in a showcase using an image, then the user changes that image, the section is not updated in the honeycomb showcase editor.
How: Changed the section to use the associated items image instead of the section image, since section.image is only ever touched when the section is first created.


jaronkk:
:+1:

jonhartzler:
:+1:

PR #219, opened 2015-08-13
fix the status message on the collections listing page

jonhartzler:
it was always showing published changed to reflect a collections actual status


jaronkk:
:+1:

jgondron:
:+1:

PR #220, opened 2015-08-14
Replace image refresh bug

jgondron:
Why: Image processing is now performed in the background. When replacing an image in the item edit form, it loads the page before processing is performed and cannot render the image.
How: Modified the ItemShowImageBox to show an indeterminate progress circle until the image is finished processing. To do this it will make 3 calls, 5 seconds apart until it fails or receives an image_ready as the image_status, at which point it will render the image as before. This required several changes, including creating an image_status field on item and changing the data that is sent from the v1 api and the chain of calls prior to this component, such as the ItemJsonDecorator.


danthewolfe:
:+1:

jaronkk:
:+1:

PR #222, opened 2015-08-17
Fix for null data from the user api

jaronkk:
Andy ran into an issue where a null values coming from the API set the user email address to null, violating the not null constraint on the user table.

We may also need to remove the not null constraint from the user table? (Or do that instead of this PR)


jonhartzler:
:+1:

jgondron:
:+1:

PR #225, opened 2015-08-19
Section image bug fix

jgondron:
Fixing a bug if the section is text only (no associated item)

### DIFF
--- a/app/views/sections/_show.json.jbuilder
+++ b/app/views/sections/_show.json.jbuilder
@@ -1,7 +1,9 @@
 json.id section.id
 json.name section.name
 json.description section.description
-json.image section.item.image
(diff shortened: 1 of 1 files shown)


jaronkk:
:+1:

app/views/sections/_show.json.jbuilder
```diff
@@ -1,7 +1,9 @@
 json.id section.id
 json.name section.name
 json.description section.description
-json.image section.item.image
+if section.item
```
jonhartzler: I think I would like to see a method added to SectionJSONDecorator to do this logic.  Then it can be tested as well.

jonhartzler:
:+1:

PR #226, opened 2015-08-19
DEC-418 Drop down suppressed appropriately

rfox2:
What: Suppressed the metadata field drop down when all metadata fields have been selected or filled in for an item.
How: Modified the ItemMetaDataForm component so that the additionalFields method tracks the number of displayed fields against the total additional field count and suppresses the drop down if they match.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: ruby
 sudo: false
 rvm:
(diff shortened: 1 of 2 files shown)


danthewolfe:
:+1:

jaronkk:
:+1:

PR #230, opened 2015-09-09
Item form marked as dirty when it should be clean

jaronkk:
Fixed issue with form being marked as dirty whenever onblur was triggered on stringfields.

The only change other than adding newlines was removing `onBlur={this.handleChange}`


jonhartzler:
:+1:

danthewolfe:
:+1:

PR #231, opened 2015-09-09
Showcase text only section layout has width for image, looks bad.

danthewolfe:
Don't render empty img tag when there is no image, render null instead.


jaronkk:
:+1:

PR #232, opened 2015-09-10
Fix display of the add items dialog

jaronkk:
There were a couple problems with the add items dialog.

1) The nav header was appearing on top of the dialog overlay.  This was fixed by a z-index tweak.
2) The dialog did not resize to fit the window size, and if too much content was present inside the dialog, it would overflow below the bottom of the window.

The second issue led me down a long and sordid path involving an update to material-ui.  The update fixed the sizing issue, but I also had to add a new form component because nested refs stopped working at some point. (The ReactDropzone component could no longer access `this.refs.uploadForm`).

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -27,4 +27,4 @@
 //= require menu-bar
 //= require flash
 //= require jquery.cookie
-$ = jQuery = require('jquery');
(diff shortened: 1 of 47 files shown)


jonhartzler:
:+1:

jgondron:
:+1:

PR #234, opened 2015-09-10
Image displays with error

jgondron:
Why: Fix for https://docs.google.com/document/d/19a97aJPiSRvBgxIGKkIkcpViOg5_Iou3OvUgcgHItXM/edit#bookmark=id.g8xtpr1ochgj
How: Problem was that it was firing off an ajax request and testing the request count at the same time. Moved the count test to only happen when we get a processing response from the server. Added an additional guard to prevent firing another request when the previous is still awaiting a response. This could theoretically cause the wait to be longer than 15 seconds if the server takes longer than 5s to respond on any one request.


jonhartzler:
:+1:

jaronkk:
:+1:

PR #238, opened 2015-09-15
Dec 498 google import

jonhartzler:


app/controllers/import_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+        sheet: params[:sheet_name] }
+    )
+    render json: { auth_uri: authorization_uri }
+  end
+
+  # This should get called whenever google redirects after a successful authorization request.
+  # We'll get an authorization code in the params and use this to make the connection
+  # to google to read the sheet data
+  def import_google_sheet_callback
+    state_hash = JSON.parse(Base64::decode64(params[:state]))
```
jonhartzler: I would be inclined to see this work in its own class.  maybe a service class.  May not need to happen now but as a refactor going forward.

app/services/create_items.rb
```diff
@@ -0,0 +1,24 @@
+# Performs batch creation of items from a json
+class CreateItems
+  def self.call(collection_id:, items_hash:, rewrite_rules: nil)
+    new.create!(collection_id: collection_id, items_hash: items_hash, rewrite_rules: rewrite_rules)
+  end
+
+  def create!(collection_id:, items_hash:, rewrite_rules: nil)
+    ActiveRecord::Base.transaction do
+      items_hash.each do |item_props|
```
jonhartzler: I would consider yielding a block here. 

Maybe: 
items_hash.each do |item_props|
  item = Item.new(item_props)
  item.collection_id = collection_id

  block.call(item) if block_given? 

  item.save  
end 



CreateItems.create(item_props) do | item | 

  rule1.rewrite(item)
  rule2.rewrite(item)

end

app/views/items/_action_bar.html.erb
```diff
@@ -1,12 +1,16 @@
+<script type="text/javascript" src="https://apis.google.com/js/api.js"></script>
```
jonhartzler: In the future we might refactor this to try this method to encapsulate the call the api code in side the react component.  

http://stackoverflow.com/questions/27530080/dynamically-insert-script-tag-that-will-execute-via-reactjs

Or we could put it in the header.

jonhartzler:
:+1:   I don't think we need to make any of the suggested changes now.  They were more observations.

app/assets/javascripts/mixins/GooglePickerMixin.jsx
```diff
@@ -0,0 +1,92 @@
...
+    this.setState({
+      pickerApiLoaded: true
+    });
+    this.createPicker();
+  },
+
+  handleAuthResult: function(authResult) {
+    if (authResult && !authResult.error) {
+      this.setState({ oauthToken: authResult.access_token });
+      this.createPicker();
```
jaronkk: There's a potential for a problem here where `this.state.oauthToken` is not set by the time `this.createPicker()` is run.  setState does not immediately update the values in this.state.

Instead, we should implement [componentDidUpdate](https://facebook.github.io/react/docs/component-specs.html#updating-componentdidupdate), and within componentDidUpdate call the createPicker method.

createPicker will also need to track a state variable of whether or not it has been initialized

```js
  createPicker: function() {
    if (this.state.pickerApiLoaded && this.state.oauthToken && !this.state.pickerInitialized) {
      this.setState({ pickerInitialized: true });
      ... 
    }
  },
```

app/controllers/import_controller.rb
jgondron: Moved most of this into the GoogleCreateItems service

jaronkk:
:+1:

PR #239, opened 2015-09-15
Travis CI notifications in Hipchat

jaronkk:
## Travis CI
This adds travis notifications in our Hipchat room.  It will only notify when the build status changes.
I also moved the database migrate from `scripts` to `before_script`.  The reason for this is because travis will collapse the output from anything run in `before_script`.  Anything run in `scripts` won't be collapsed, and the migrate task is very verbose, leading to builds that are hard to read.

## Spec fix
I also increased the time given for allowing Solr to start, since the original 10 seconds sometimes was not enough time for Travis to start Solr.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,11 @@ rvm:
 before_install:
   - npm install
 script:
-  - RAILS_ENV=test bundle exec rake --trace db:migrate test
   - bundle exec rspec spec
 before_script:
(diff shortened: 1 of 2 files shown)


jonhartzler:
:+1:

rfox2:
:+1:  :+1:  :+1:  :+1:  :+1:  :+1:

jgondron:
:100: x :+1:

PR #240, opened 2015-09-16
DEC-433: Records created without a unique_id

jgondron:
Why: There was potential for records being created without a unique_id, causing lots of other problems since this is expected to be populated for all records.
How:
-Changed CreateUniqueId to no longer save the item
-Moved the generation of the id in the Save services to happen prior to the initial save
-Added validation on all of the models that have a unique_id since we never want any of these objects to be created without it


jaronkk:
:+1:

danthewolfe:
:+1:

PR #241, opened 2015-09-16
Dec 498 import migration

jgondron:
Depends on https://github.com/ndlib/honeycomb/pull/240
Why: user_defined_id is now a required field, but the default path for creating items was not generating it.
How: Changed SaveItem to copy the unique_id into the user_defined_id field if it's not defined. If it is defined, such as through import, it will use the given id. The migration will do the same for records that already exist.


jonhartzler:
:+1:

jaronkk:
:+1:

PR #242, opened 2015-09-16
Dec 566 add api end point for metadata configuration.

jonhartzler:
First pull request for just the api endpoint of the metadata configuration.

### DIFF
--- a/app/models/metadata/configuration.rb
+++ b/app/models/metadata/configuration.rb
@@ -7,7 +7,6 @@ module Metadata
       @item_configuration ||= new(load_yml(:item))
     end
 
-#    private_class_method :load_yml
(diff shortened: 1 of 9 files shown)


app/models/metadata/configuration.rb
```diff
@@ -37,10 +33,28 @@ def label?(name)
       label(name).present?
     end
 
-    private_class_method :load_yml
+    def to_json(_options = {})
```
jaronkk: Can you break this into two methods? An as_json method that returns the hash, and a to_json method that just calls as_json.to_json?

jaronkk:
:+1:

jgondron:
:+1:

PR #243, opened 2015-09-17
DEC-574 fix redactor HTMLField component

jaronkk:
Editing the source in an HTMLField component was yielding strange behaviors because the code was not updating the props with the raw user input.  Changed to use the raw user input in the onChange event on the textarea, and added separate functions for the callbacks from Redactor vs the textarea.

redactor_field.js.coffee: buttonSource is now just source in newer versions of Redactor.


jonhartzler:
:+1:

jgondron:
:+1:

PR #244, opened 2015-09-17
DEC-529: import validation

jgondron:
Why: Need to validate all records and report back to the user meaningful information about the import
How: Refactored GoogleCreateItems and ImportController to provide more data about the valid/invalid records imported. GoogleCreateItems will now return a json with a summary and a list of errors. Created a basic view to show the data. This will likely move to a react component that can be displayed without leaving the items page (DEC-575) so I didn't spend a whole lot of time on its appearance.

Example of the results view:

screen shot 2015-09-17 at 3 09 23 pm

### DIFF
--- a/app/services/google_create_items.rb
+++ b/app/services/google_create_items.rb
@@ -2,6 +2,7 @@
 # Assumes you have a valid auth token to access the doc
 class GoogleCreateItems
   attr_reader :session
+  attr_reader :total_count, :valid_count, :new_count, :error_count, :changed_count
 
   # Simplified call to create_from_worksheet!
   def self.call(auth_code:, callback_uri:, collection_id:, file:, sheet:)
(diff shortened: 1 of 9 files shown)


app/services/google_create_items.rb
```diff
@@ -34,16 +45,44 @@ def create_from_worksheet!(collection_id:, file:, sheet:)
...
+      errors_array.compact
+    end
+  end
+
+  def reset_counts
+    @total_count = 0
+    @valid_count = 0
+    @new_count = 0
+    @error_count = 0
+    @changed_count = 0
```
jaronkk: Rather than using a method like this, what do you think about initializing these value in the initialize method?

jgondron: Well this is in the off chance that a single instance ever gets used to import more than once. They were just local variables in the create! method, but I moved them out due to ABC warnings. If you guys are ok with the ABC warnings I'll move them back in. I would prefer to tbh as it's better encapsulated since create! is the only thing that ever uses them.

jaronkk:
:+1:

jaronkk:
:+1:

jonhartzler:
:+1:

PR #246, opened 2015-09-24
Dec 555 external collections

rfox2:
What: Added a small workflow for creating and maintaining external collections
How: Several react components were added, a new composite model was added, as well as a new controller. There are several changes to some of the pre-existing services. A URL field was added to the Exhibit model that we can use for redirects to the external sites.


app/assets/javascripts/components/forms/ButtonRaised.jsx
```diff
@@ -0,0 +1,38 @@
+/** @jsx React.DOM */
+var React = require('react');
+var mui = require("material-ui");
+var RaisedButton = mui.RaisedButton;
+var FontIcon = mui.FontIcon;
+
+var ButtonRaised = React.createClass({
```
jonhartzler: I think I would like to see this class have a slightly more descriptive name.  It looks like it goes to a link.  So maybe ButtonLink or it is an action bar button link.

rfox2: Changed name to "ButtonLink"

jgondron:
:+1:

jonhartzler:
:+1:

PR #247, opened 2015-09-25
Dec 566 add metadata confg to forms

jonhartzler:
Added the configuration from the api to the item meta data form.


config/metadata/item.yml
```diff
@@ -85,6 +89,7 @@
   :default_form_field: false
   :optional_form_field: true
   :order: 12
+  :placeholder: "Example \"Copyright held by Hesburgh Libraries\""
```
jaronkk: Add a colon after "Example" for consistency.

jaronkk:
:+1: after my comment

app/assets/javascripts/stores/MetaDataConfigurationStore.js
```diff
@@ -0,0 +1,38 @@
...
+    AppDispatcher.register(this.receiveAction.bind(this));
+  }
+
+  receiveAction() {
+    // held for use when actions are sent back to the object.
+  }
+
+  getAll(succesFunction) {
+    var id = CollectionStore.uniqueId;
+    var url = "/v1/collections/" + id + "/metadata_configuration";
```
jgondron: Any (good) way that we can pass this in from the rails app? Not sure how easy that would be since this one isn't using an initializer.

jonhartzler: There will be another ticket to update the app to load the collection off the api. once that is done the url for the configuration will be embedded in the collection data and we can use that to get the configuration.

jgondron:
:+1:

PR #248, opened 2015-09-25
DEC-499: Update items on import

jgondron:
Why: Users would like imports to update existing items, not just create new ones
How: 
-Updated the services that create the items to do a find or create by user defined id and collection id. This meant having to also add a unique constraint on collection_id, user_defined_id in items since we are relying on this to uniquely identify items for imports. 
-Added an unchanged count to the summary to inform the user how many items matched up to existing records but were not changed.

### DIFF
--- a/app/models/metadata/configuration.rb
+++ b/app/models/metadata/configuration.rb
@@ -33,6 +33,10 @@ module Metadata
       label(name).present?
     end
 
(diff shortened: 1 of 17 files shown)


jaronkk:
:+1:

rfox2:
:+1:

PR #249, opened 2015-09-25
Add search functionality via Solr!

jaronkk:
This adds search functionality with keyword searching, facets, and sorting.

DEC-401, DEC-506, DEC-516, DEC-517, DEC-518, DEC-557


app/models/metadata/configuration.rb
```diff
@@ -33,10 +31,71 @@ def label?(name)
...
+
+    def sort_map
+      @sort_map ||= build_sort_map
+    end
+
+    def build_fields
+      data.fetch(:fields).map { |field_data| self.class::Field.new(**field_data) }
+    end
+
+    def build_facets
```
jonhartzler: Do you want to @var ||=  this ?

jaronkk: The facets method handles caching the result

app/models/waggle/adapters/solr/search/hit.rb
```diff
@@ -0,0 +1,74 @@
...
+          def fetch_datetime(field)
+            value = fetch(Waggle::Adapters::Solr::Types::DateTime.field_name(field))
+            if value.present?
+              Waggle::Adapters::Solr::Types::DateTime.from_solr(value)
+            end
+          end
+
+          def fetch(field)
+            value = solr_doc.fetch(field.to_s, nil)
+            if value.is_a?(Array)
```
jonhartzler: Why is it ok to return the first result only in an array ?

app/decorators/v1/item_json_decorator.rb
```diff
@@ -39,10 +37,40 @@ def metadata
...
+      json.set! "isPartOf/collection", collection_url
+      json.id unique_id
+      json.collection_id collection_id
+      json.slug slug
+      json.name name
+      json.description description.to_s
+      json.image image
+      json.metadata metadata
+      json.last_updated updated_at
+    end
```
jgondron: I think you need this after the method:
# rubocop:enable Metrics/AbcSize

app/models/waggle/adapters/solr/search/hit.rb
jaronkk: The way we described the search API we have it returning single values for all displayed fields.  This particular case is because we're returning the name_t value for the name, which is a multivalued field even though we're expecting only one value to ever be stored for name.  All of the other fields used by Hit are indexed as strings or datetimes and are single value.

The alternative to returning the first value if it's an Array would be to specifically index a name_s so we know it's not multivalued.

app/decorators/v1/item_json_decorator.rb
jaronkk: If you put the rubycop:disable functionality on the same line as a method definition it will only be disabled for that particular method.

app/models/waggle/search/result.rb
```diff
@@ -8,7 +8,7 @@ def initialize(query: query)
       end
 
       def hits
```
jonhartzler: This is probably not for now but more of a comment for posterity.  It looks like the Adapter::Result also runs an array map.  It might be nice if the there was only one space that results got mapped to classes.  Maybe if you always generated Search::Hit and that took care of mapping to the adapter specific hit.

Gemfile
```diff
@@ -23,6 +23,7 @@ gem "fastimage"
 
 gem "hesburgh_infrastructure", github: "ndlib/hesburgh_infrastructure"
 gem "hesburgh_api", github: "ndlib/hesburgh_api"
+gem "honeycomb_solr", github: "ndlib/honeycomb_solr"
```
rfox2: Is this generic enough to name it something other than honeycomb_*? Could we make use of this in other applications that use solr?

app/models/waggle/adapters/solr.rb
```diff
@@ -0,0 +1,36 @@
+module Waggle
+  module Adapters
+    module Solr
+      def self.setup
```
rfox2: Why the empty method?

app/models/waggle/adapters/solr/index.rb
```diff
@@ -0,0 +1,39 @@
+module Waggle
+  module Adapters
+    module Solr
+      module Index
+        SUFFIXES = {
```
rfox2: Should this be parameterized in a yaml file?

app/models/waggle/adapters/solr/search/result.rb
```diff
@@ -0,0 +1,155 @@
+module Waggle
+  module Adapters
+    module Solr
+      module Search
+        class Result
+          FIELD_BOOSTS = {
```
jgondron: Can this be specified and read from the configuration?

jgondron:
:+1: after comments are addressed and we've tested the beehive branch against this.

app/models/waggle/adapters/solr/index/item.rb
```diff
@@ -0,0 +1,79 @@
...
+          def id
+            "#{waggle_item.id} #{waggle_item.type}"
+          end
+
+          def metadata
+            @metadata ||= Waggle::Adapters::Solr::Index::Metadata.new(metadata_set: waggle_item.metadata)
+          end
+
+          def as_solr
+            @as_solr ||= metadata.as_solr.clone.tap do |hash|
```
rfox2: Out of curiosity, is this being cloned because the returned hash is frozen?

app/models/waggle/adapters/solr/search/hit.rb
rfox2: I like your option #2 a little better. The way this is written it seems a little ambiguous.

config/metadata/item.yml
```diff
@@ -1,123 +1,139 @@
...
-  :placeholder: "Example: \"Ballantine Books\""
-- :name: :manuscript_url
-  :type: :string
-  :label: Digitized Manuscript
-  :default_form_field: false
-  :optional_form_field: true
-  :order: 16
-  :placeholder: "http://"
-  :help: Link to externally hosted manuscript viewer.
+:fields:
```
rfox2: Is there any reason why we couldn't put field weighting in this config file as well?

rfox2:
:+1:  Added some comments and questions but following that, I'm good

app/models/waggle/adapters/solr/index.rb
jaronkk: It could be, but I felt it was small enough to manage as a constant.  I'm not particularly tied to either methodd

config/metadata/item.yml
jaronkk: Done

app/models/waggle/adapters/solr/search/result.rb
jaronkk: Done

Gemfile
jaronkk: It probably could be.  It's essentially a wrapper around jettywrapper, which is a wrapper around jetty, which serves Solr. Whee!

PR #251, opened 2015-09-29
DEC-433: Add Item Bug

jgondron:
Why: We are currently unable to add more than one new item through the add items form in HoneyComb.
How:
-Problem was due to the fact that CreateUniqueId required an id for the object, which required saving the object to the database prior to calling this service class. Changed it to use SecureRandom.hex to generate a unique (enough) id, removing this dependence.
-Also changed the initial population of user_defined_id to be independent of the unique_id. It now has its own service class that generates an id using SecureRandom.hex. This one doesn't require as complex an id because it just needs to be unique within a collection.
-Also fixed a stubbed out test was that never implemented within create_items_spec and changed a few tests in save_item_spec to be more unit testish.

### DIFF
--- a/app/services/create_unique_id.rb
+++ b/app/services/create_unique_id.rb
@@ -23,7 +23,7 @@ class CreateUniqueId
   private
 
   def unique_id
-    SecureRandom.uuid
+    SecureRandom.hex(5)
   end
 
   def validate_interface!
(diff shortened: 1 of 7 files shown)


jonhartzler:
:+1:

rfox2:
:+1:

PR #253, opened 2015-09-29
changed the default search response rows.

jonhartzler:
12 makes a better default number because it breaks even if there are 4, 3 and 2 results per line.


rfox2:
:+1:

PR #255, opened 2015-10-01
fix bug with ruby's iso8601 parser.

jonhartzler:
Also change the metadata date indexer to not convert an iso date to a
date so it can be converted back to iso to be indexed.


jgondron:
:+1:

danthewolfe:
:+1:

PR #256, opened 2015-10-01
DEC-497: Export collection

jgondron:
Why: User needs a way to export a collection to a Google spreadsheet in the correct format
How: Added client side code that creates the spreadsheet then requests an auth token for access to the new sheet that the server can use to populate the data. Added routes/methods server side to export item data for a collection using the auth token and worksheet provided by the client. Once the export is complete the server will redirect the user to the new sheet.

Note: The google api does not currently allow programmatically setting a cells format. Because of this, by default, it will attempt to translate dates into a new format which will cause problems when reimporting. For the time being I'm using the old string literal trick (which google supports in its sheets) by prepending the value with a ' character. I don't *really* like this solution because it's known to cause problems for users who want to do additional processing with their data. In the future, we may want to just separate the date fields into separate columns, but this would be a fairly big change to both export and import so I want to try this simple method first.

### DIFF
--- a/app/assets/javascripts/mixins/GoogleCreatorMixin.jsx
+++ b/app/assets/javascripts/mixins/GoogleCreatorMixin.jsx
@@ -1,0 +1,76 @@
+// Requires the following within the page that uses this mixin:
+//  <script type="text/javascript" src="https://apis.google.com/js/api.js"></script>
+//
+
+var GoogleCreatorMixin = {
+  propTypes: {
+    developerKey: React.PropTypes.string.isRequired,
+    clientId: React.PropTypes.string.isRequired,
+    appId: React.PropTypes.string.isRequired,
+    authUri: React.PropTypes.string.isRequired,
+  },
+
+  oauthToken: null,
+
+  clientApiLoaded: false,
+
+  loadCreator: function() {
+    if(this.clientApiLoaded && this.oauthToken){
+      this.createFile(this.fileCreated);
+    } else {
+      gapi.load("auth", {"callback": this.onAuthApiLoad});
+      gapi.load("client", {"callback": this.onClientApiLoad});
+    }
+  },
+
+  onClientApiLoad: function() {
+    this.clientApiLoaded = true;
+    this.createFile(this.fileCreated);
+  },
+
+  onAuthApiLoad: function() {
+    gapi.auth.init(this.onAuthApiInit);
+  },
+
+  onAuthApiInit: function(clientId) {
+    window.gapi.auth.authorize(
+      {
+        "client_id": this.props.clientId,
+        "scope": "https://www.googleapis.com/auth/drive.file",
+        "hd": "nd.edu"
+      },
+      this.handleAuthResult);
+  },
+
+  handleAuthResult: function(authResult) {
+    if (authResult && !authResult.error) {
+      this.oauthToken = authResult.access_token;
+      this.createFile(this.fileCreated);
+    }
+  },
+
+  createFile: function(callback) {
+    if(this.clientApiLoaded && this.oauthToken){
+      const boundary = 'foo_bar_baz';
+      const delimiter = "\r\n--" + boundary + "\r\n";
+      const close_delim = "\r\n--" + boundary + "--";
+
+      var fileData = this.getFileData();
+
+      var metadata = {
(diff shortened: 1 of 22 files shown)


app/assets/javascripts/mixins/GoogleCreatorMixin.jsx
```diff
@@ -0,0 +1,76 @@
...
+  handleAuthResult: function(authResult) {
+    if (authResult && !authResult.error) {
+      this.oauthToken = authResult.access_token;
+      this.createFile(this.fileCreated);
+    }
+  },
+
+  createFile: function(callback) {
+    if(this.clientApiLoaded && this.oauthToken){
+      const boundary = 'foo_bar_baz';
```
rfox2: Nice. :-)

app/controllers/export_controller.rb
```diff
@@ -0,0 +1,40 @@
+class ExportController < ApplicationController
+  # Constructs an auth request to google. Packs the collection id, file, and sheet
+  # into state data so that this data persists the round trip.
+  def get_google_export_authorization_uri
```
rfox2: What happens down the line if an invalid collection id is submitted here?

app/services/google_export_items.rb
```diff
@@ -0,0 +1,40 @@
...
+    instance = new(auth_code: auth_code, callback_uri: callback_uri)
+    instance.export_to_worksheet!(items: items, file: file, sheet: sheet)
+  end
+
+  def initialize(auth_code:, callback_uri:)
+    @session = GoogleSession.new
+    session.connect(auth_code: auth_code, callback_uri: callback_uri)
+  end
+
+  def export_to_worksheet!(items:, file:, sheet:)
```
rfox2: Just curious: why the bang?

app/services/google_session.rb
```diff
@@ -63,4 +63,33 @@ def worksheet_to_hash(worksheet:)
...
+      # See if this hash has any new columns. If so, append it to the end
+      hash.keys.each do |key|
+        unless header_row.include?(key)
+          header_row << key
+        end
+      end
+
+      # Populate the row with values based on the order of the header row
+      row = []
+      header_row.each do |header|
```
rfox2: Does this mean that each row of the resulting spreadsheet will be an array of values from all of the records?

jgondron: To assign values to a row you just give it a 1d array of values, where the index in the array is the column that it will map to. So, it essentially has to map nonuniform data to a 2d array since some items may have different metadata keys in the hash. So it has to figure out the total number of columns to populate, then make sure all items map their values correctly to those columns. Items that did not have a certain key will just leave that field blank in the array. Not sure if this answers the question.

app/services/google_export_items.rb
jgondron: No strong reason. I think because it modifies the object. In this case that should be pretty clear that will happen even without it.

app/controllers/export_controller.rb
jgondron: At this point it isn't doing anything with any of our objects. It's only forming a request uri that the client can use to request a token that the server can then use to r/w to a file in google docs. The export call back will eventually do a check for permissions to edit the collection, which should fail if it doesn't exist. So it will waste a round trip from server, to client, to gapi, then back to our server, but nothing damaging will happen as far as I can tell.

app/assets/javascripts/mixins/GoogleCreatorMixin.jsx
jgondron: Ah, I actually meant to change that :)

app/services/google_export_items.rb
rfox2: OK. It wasn't clear to me how the object was modified but I trust you that it is. :-)

app/services/google_session.rb
rfox2: Right, I just wasn't sure if the tuples would be vertically aligned or horizontally aligned. With the header row dictating the values of the column, I think that answers my question. I was just making sure that the tuples were horizontally aligned but they have to be in order for this to work I think. Ignore the question. :-)

rfox2:
:+1:

jonhartzler:
:+1:

PR #257, opened 2015-10-02
DEC-587, DEC-556 Add external url value to V1 output

rfox2:
What: Added field to API output for external collection urls


jgondron:
:+1:

danthewolfe:
:+1:

app/decorators/v1/collection_json_decorator.rb
```diff
@@ -23,6 +23,10 @@ def metadata_configuration_url
       h.v1_collection_metadata_configuration_url(object.unique_id)
     end
 
+    def external_url
```
jonhartzler: There is a spec for this file.  collection_json_decorator_spec.rb.  I think I would like to see a spec on this method.

rfox2: My bad. I was looking at the wrong collection_json_decorator_spec.rb file and only saw one expectation in it, so it looked like an all or nothing proposition. I'll add in this spec when I push up the other changes.

PR #261, opened 2015-10-05
Add additional type designation in V1 output for collections

rfox2:
**What** Added additional type designation for collections output via the V1 API. Schema.org allows for an "additionalType" field which is typed as a URL. I also added definitions for the two new sub-types on the honeycomb wiki. I modified the parent context for collections to be "bib.schema.org" and changed the type to "Collection".

**Why** This gives the API consumer a way to easily identify what type of collection is being output. The "Collection" type, which is a sub-type of "CreativeWork" is also more appropriate for the type of things being output via the API in aggregate.

**What** Added a new method to the collection decorator, unit tests, and modified the jbuilder file as outlined above.

### DIFF
--- a/app/decorators/v1/collection_json_decorator.rb
+++ b/app/decorators/v1/collection_json_decorator.rb
@@ -27,6 +27,10 @@ module V1
       object.exhibit.url ? object.exhibit.url : ""
     end
 
+    def additional_type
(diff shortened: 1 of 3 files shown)


jonhartzler:
:+1:

jgondron:
:+1:

PR #263, opened 2015-10-06
Import incorrectly showing updates

jgondron:
Why: There is a difference between the item edit form and import in how they treat unset/empty metadata. This is causing import to report record changes when, from the user's perspective, there was no changes.
How:
-Modified item decorator to only send a metadata field if it actually has content. This is to prevent frontends from processing and sending update data for fields that actually have not been set.
-To catch any additional problems in the attribute data sent to SaveItem, added a ParamCleaner that does some common cleanup such as string bools to bools and empty string to nil

### DIFF
--- a/app/decorators/item_decorator.rb
+++ b/app/decorators/item_decorator.rb
@@ -60,24 +60,7 @@ class ItemDecorator < Draper::Decorator
 
   def meta_data
     data = {}
-    [
-      :name,
-      :description,
-      :transcription,
-      :manuscript_url,
-      :creator,
-      :contributor,
-      :subject,
-      :publisher,
-      :alternate_name,
-      :rights,
-      :call_number,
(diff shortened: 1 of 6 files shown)


jonhartzler:
:+1:

app/decorators/item_decorator.rb
```diff
@@ -55,24 +55,35 @@ def item_meta_data_form
...
-        date_created: object.date_created,
-        date_published: object.date_published,
-        date_modified: object.date_modified,
-      })
+      data: meta_data)
+  end
+
+  def meta_data
+    data = {}
+    [
```
jonhartzler: can we read this out of configuration?  

Configuration#field_names  maybe.

rfox2:
:+1:  once the tests are fixed

PR #264, opened 2015-10-06
Update item image status to have a no image state.

jonhartzler:
Why:  this is necessary so we can display correctly in the UI when an Item does not have an image associated with it. 

What:  I added a new status to the image status.  Changed where that status gets set in save item.  Updated the listing page and item edit form to look at the status.


rfox2:
:+1: the jsx looks a lot cleaner now; much more legible

app/assets/javascripts/components/ItemShowImageBox.jsx
```diff
@@ -27,37 +22,12 @@ var ItemShowImageBox = React.createClass({
...
-    {
-      case "image_ready":
-        this.setState({
-          imageReady: true,
-          image: image["items"]["image"],
-        });
-        clearInterval(this.state.requestTimer);
-        break;
-      case "image_processing":
-        if(this.state.requestCount >= this.props.maxRetries) {
```
jgondron: Did you mean to remove this bit? This prevented the client from infinitely pinging the server.

app/assets/javascripts/components/ItemShowImageBox.jsx
```diff
@@ -67,60 +37,78 @@ var ItemShowImageBox = React.createClass({
       requestCount: this.state.requestCount + 1
     });
 
+    console.log(this.props.itemPath);
```
jgondron: ^

app/assets/javascripts/components/ItemShowImageBox.jsx
jonhartzler: I did.  Basically I looked at it and said.  Normal operation has this as not a common occurrence.  One request every 3 seconds is not a load altering request especially since the result is a cache hit until the item changes.  I decided not to pre-optimize and add complexity that does not provide immediate benefit.

jgondron: Well, load was a concern, but the biggest reason was ux. I didn't want the user to think they'd had to wait indefinitely for the image to load to know that it was successful (especially within the context of replacing an image). I wanted them to know they could leave the page and continue working and come back if they needed to. So, I wanted to be able to reliably say that the user would only ever wait a max of 15 seconds before knowing that things are probably ok, and they can come back later.

PR #266, opened 2015-10-08
Import images script

jgondron:
Why: Need a way to import images in bulk to support one-off collections that need to be created prior to supporting bulk image import through the interface.
How: Created a temporary rake task that we can use in the interim. It expects a collection id and a single text file with paths to the files that need to be imported on separate lines. It does not handle traversing directories to magically pull files in, so any collating of files will need to happen prior to the import when generating the text file.


danthewolfe:
:+1:

rfox2:
:+1:

PR #267, opened 2015-10-09
DEC-616 Migration of url field to collection model

rfox2:
What: Schema, code and data migration of URL field from the exhibit model to the collection model
How: Code, schema and data migrated


jgondron:
:+1:

jonhartzler:
:+1:

PR #268, opened 2015-10-12
DEC-615: Merge exhibit name into collection name_line_1, name_line_2

jgondron:
There don't seem to be any references to exhibit.name other than exhibition. Most objects are already using collection.name_line_1 and name_line_2.


jonhartzler:
:+1:

rfox2:
:+1:

PR #269, opened 2015-10-12
DEC-625: Migrate about text to collections

jgondron:
-Created migration to add about text column to Collections and copy existing data from the exhibits.about into collections.about
-Changed all references to exhibits.about to collections.about


rfox2:
Make PR against DEC-376-Merge-collection-and-exhibit-models

jonhartzler:
:+1:

PR #270, opened 2015-10-12
DEC-629: Migrate copyright

jgondron:
-Created migration to add copyright text column to Collections and copy existing data from the exhibits.copyright into collections.copyright
-Changed all references to exhibits.copyright to collections.copyright


rfox2:
Make PR against DEC-376-Merge-collection-and-exhibit-models

PR #271, opened 2015-10-12
DEC-630: Migrate enable_browse

jgondron:
-Created migrations to add enable_browse text column to Collections and copy existing data from the exhibits. enable_browse into collections. enable_browse
-Changed all references to exhibits. enable_browse to collections. enable_browse


rfox2:
Make new PR against DEC-376-Merge-collection-and-exhibit-models

PR #272, opened 2015-10-13
DEC-617 Merge description fields into collection model

rfox2:
**What**: Two description fields have been merged into the collection model and renamed to site_intro and short_intro to be consistent with how we're using them in the application.
**How**: Two fields were added to the collection model and the API was updated.

### DIFF
--- a/db/migrate/20151009171750_migrate_descriptions_to_collections.rb
+++ b/db/migrate/20151009171750_migrate_descriptions_to_collections.rb
@@ -1,0 +1,26 @@
+class MigrateDescriptionsToCollections < ActiveRecord::Migration
+  class Exhibit < ActiveRecord::Base
+  end
+
+  def up
+    add_column :collections, :site_intro, :text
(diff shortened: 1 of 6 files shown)


db/migrate/20151009171750_migrate_descriptions_to_collections.rb
```diff
@@ -0,0 +1,26 @@
...
+  class Exhibit < ActiveRecord::Base
+  end
+
+  def up
+    add_column :collections, :site_intro, :text
+    add_column :collections, :short_intro, :text
+    Exhibit.find_each do |exhibit|
+      site_intro = exhibit.description
+      short_intro = exhibit.short_description
+      execute "UPDATE collections SET site_intro = '#{site_intro}' WHERE id = #{exhibit.collection_id}" unless site_intro.nil?
```
jgondron: Can you change to do this in one statement?

PR #273, opened 2015-10-13
DEC-617 Merge description fields into collection model

rfox2:
**What**: Two description fields have been merged into the collection model and renamed to site_intro and short_intro to be consistent with how we're using them in the application.

**How**: Two fields were added to the collection model and the API was updated.


jonhartzler:
:+1:

jgondron:
:+1:

PR #277, opened 2015-10-13
DEC-633 Merge image fields into collections

rfox2:
What: Integrated all image file references from exhibits into the collections model. Included in this merge is the honeypot image relationship.
How: Added migrations for both the exhibits image fields and the honeypot image model. The exhibition model was also updated. The save collection service class was updated, but I didn't touch the save exhibit service class so cleanup will need to happen when we remove the exhibit class entirely.

### DIFF
--- a/app/models/exhibition.rb
+++ b/app/models/exhibition.rb
@@ -13,21 +13,21 @@ class Exhibition
     :unique_id,
     :exhibit,
     :description,
+    :about,
     :published,
     :preview_mode,
+    :copyright,
     :short_intro,
     :site_intro,
     :updated_at,
     :image,
     :honeypot_image,
     :uploaded_image,
+    :hide_title_on_home_page,
+    :showcases,
     :url
   ]
   exhibit_methods = [
-    :showcases,
-    :about,
-    :copyright,
-    :hide_title_on_home_page,
   ]
 
   delegate *collection_methods, to: :collection
--- a/app/models/exhibition.rb
+++ b/app/models/exhibition.rb
@@ -18,12 +18,12 @@ class Exhibition
     :short_intro,
     :site_intro,
     :updated_at,
-    :url
-  ]
-  exhibit_methods = [
     :image,
     :honeypot_image,
     :uploaded_image,
+    :url
+  ]
+  exhibit_methods = [
     :showcases,
     :about,
     :copyright,
@@ -73,14 +73,14 @@ class Exhibition
   # rubocop:disable Metrics/AbcSize
   def save!
     ActiveRecord::Base.transaction do
-      SaveCollection.call(collection, collection.as_json)
       exhibit.collection = collection
-      if exhibit.uploaded_image.size
-        exhibit_attrs = exhibit.as_json.merge(uploaded_image: exhibit.uploaded_image)
+      if collection.uploaded_image.size
+        collection_attrs = collection.as_json.merge(uploaded_image: collection.uploaded_image)
       else
-        exhibit_attrs = exhibit.as_json
+        collection_attrs = collection.as_json
       end
-      SaveExhibit.call(exhibit, exhibit_attrs)
+      SaveCollection.call(collection, collection_attrs)
+      SaveExhibit.call(exhibit, exhibit.as_json)
     end
   end
   # rubocop:enable Metrics/AbcSize
--- a/db/migrate/20151013151133_migrate_images_to_collections.rb
+++ b/db/migrate/20151013151133_migrate_images_to_collections.rb
@@ -1,0 +1,42 @@
+class MigrateImagesToCollections < ActiveRecord::Migration
+  class Exhibit < ActiveRecord::Base
+  end
+
+  def up # rubocop:disable Metrics/AbcSize
+    add_column :collections, :image_file_name, :string
+    add_column :collections, :image_content_type, :string
+    add_column :collections, :image_file_size, :integer
+    add_column :collections, :image_updated_at, :datetime
+    add_column :collections, :uploaded_file_name, :string
+    add_column :collections, :uploaded_image_content_type, :string
+    add_column :collections, :uploaded_image_file_size, :integer
+    add_column :collections, :uploaded_image_updated_at, :datetime
+    Exhibit.find_each do |exhibit|
+      file_name = exhibit.image_file_name
+      content_type = exhibit.image_content_type
+      file_size = exhibit.image_file_size
+      updated_at = exhibit.image_updated_at.utc.to_s(:db)
+      execute "UPDATE collections SET image_file_name = '#{file_name}', \
+image_content_type = '#{content_type}', \
+image_file_size = #{file_size}, \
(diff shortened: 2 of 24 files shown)


db/migrate/20151013151133_migrate_images_to_collections.rb
```diff
@@ -0,0 +1,42 @@
+class MigrateImagesToCollections < ActiveRecord::Migration
+  class Exhibit < ActiveRecord::Base
+  end
+
+  def up # rubocop:disable Metrics/AbcSize
```
rfox2: Not sure why this wasn't disabled

app/models/exhibition.rb
```diff
@@ -73,14 +73,14 @@ def collection
   # rubocop:disable Metrics/AbcSize
   def save!
     ActiveRecord::Base.transaction do
-      SaveCollection.call(collection, collection.as_json)
       exhibit.collection = collection
-      if exhibit.uploaded_image.size
-        exhibit_attrs = exhibit.as_json.merge(uploaded_image: exhibit.uploaded_image)
+      if collection.uploaded_image.size
```
jgondron: Why is this needed and where will this take place once Exhibition/Exhibit are gone?

db/migrate/20151013151133_migrate_images_to_collections.rb
```diff
@@ -0,0 +1,42 @@
...
+    add_column :collections, :uploaded_file_name, :string
+    add_column :collections, :uploaded_image_content_type, :string
+    add_column :collections, :uploaded_image_file_size, :integer
+    add_column :collections, :uploaded_image_updated_at, :datetime
+    Exhibit.find_each do |exhibit|
+      file_name = exhibit.image_file_name
+      content_type = exhibit.image_content_type
+      file_size = exhibit.image_file_size
+      updated_at = exhibit.image_updated_at.utc.to_s(:db)
+      execute "UPDATE collections SET image_file_name = '#{file_name}', \
```
jgondron: Should also copy the uploaded_image data. If we have any item that failed to process for whatever reason, we can rerun the background jobs to try again, but only if this field has data. So on the off chance that we have honeypot or sneakers issues right before deploying, this migration would break those items permanently.

db/migrate/20151013151133_migrate_images_to_collections.rb
```diff
@@ -0,0 +1,42 @@
...
+image_content_type = '#{content_type}', \
+image_file_size = #{file_size}, \
+image_updated_at = '#{updated_at}' WHERE id = #{exhibit.collection_id}" unless file_name.nil?
+    end
+
+    Exhibit.reset_column_information
+    Collection.reset_column_information
```
jgondron: Is this needed?

db/migrate/20151013151133_migrate_images_to_collections.rb
```diff
@@ -0,0 +1,42 @@
...
+    remove_column :collections, :image_content_type
+    remove_column :collections, :image_file_size
+    remove_column :collections, :image_updated_at
+    remove_column :collections, :uploaded_image_file_name
+    remove_column :collections, :uploaded_image_content_type
+    remove_column :collections, :uploaded_image_file_size
+    remove_column :collections, :uploaded_image_updated_at
+
+    Exhibit.reset_column_information
+    Collection.reset_column_information
```
jgondron: Is this needed?

db/migrate/20151013165130_migrate_honeypot_image_to_collections.rb
```diff
@@ -0,0 +1,25 @@
+class MigrateHoneypotImageToCollections < ActiveRecord::Migration
+  class HoneypotImage < ActiveRecord::Base
```
jgondron: Is this needed?

db/migrate/20151013165130_migrate_honeypot_image_to_collections.rb
```diff
@@ -0,0 +1,25 @@
...
+  def up
+    add_column :honeypot_images, :collection_id, :integer
+    Exhibit.find_each do |exhibit|
+      collection_id = exhibit.collection_id
+      exhibit_id = exhibit.id
+      execute "UPDATE honeypot_images SET collection_id = #{collection_id} WHERE exhibit_id = #{exhibit_id}"
+    end
+
+    Exhibit.reset_column_information
+    Collection.reset_column_information
```
jgondron: Is this needed?

db/migrate/20151013165130_migrate_honeypot_image_to_collections.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def down
+    remove_column :honeypot_images, :collection_id
+
+    Exhibit.reset_column_information
+    Collection.reset_column_information
```
jgondron: Is this needed?

app/models/exhibition.rb
rfox2: We need this in order to trigger image processing. The check is there to see if an actual image was uploaded.

jonhartzler: I think that paperclip responds to empty? or present? to tell if there is an object.

db/migrate/20151013151133_migrate_images_to_collections.rb
jgondron: .utc will fail when updated_at is null

db/migrate/20151013151133_migrate_images_to_collections.rb
```diff
@@ -0,0 +1,42 @@
...
+    add_column :collections, :image_updated_at, :datetime
+    add_column :collections, :uploaded_file_name, :string
+    add_column :collections, :uploaded_image_content_type, :string
+    add_column :collections, :uploaded_image_file_size, :integer
+    add_column :collections, :uploaded_image_updated_at, :datetime
+    Exhibit.find_each do |exhibit|
+      file_name = exhibit.image_file_name
+      content_type = exhibit.image_content_type
+      file_size = exhibit.image_file_size
+      updated_at = exhibit.image_updated_at.utc.to_s(:db)
```
jgondron: This fails when image_updated_at is null

db/migrate/20151013151133_migrate_images_to_collections.rb
```diff
@@ -0,0 +1,42 @@
+class MigrateImagesToCollections < ActiveRecord::Migration
+  class Exhibit < ActiveRecord::Base
+  end
+
+  def up # rubocop:disable Metrics/AbcSize
+    add_column :collections, :image_file_name, :string
+    add_column :collections, :image_content_type, :string
+    add_column :collections, :image_file_size, :integer
+    add_column :collections, :image_updated_at, :datetime
+    add_column :collections, :uploaded_file_name, :string
```
jgondron: Should this be uploaded_image_file_name

jonhartzler:
:+1:

rfox2:
:+1:

PR #280, opened 2015-10-15
DEC-636: Migrate Showcase relationship

jgondron:
Why: Need to create a relationship between Collection and Showcase now that Exhibit will no longer provide that relationship.
How: 
-Added association to Collection/Showcase and removed from Exhibit. 
-Fixed all service classes that were getting to showcases/collections via the exhibit relationship, ex: Destroy methods now directly destroy showcases related to a collection instead of going through exhibits
-Fixed the controllers use collection.showcases instead of collection.exhibit.showcases

Had to change lots of specs in this one, so please review this carefully.


jonhartzler:
:+1:

app/controllers/admin/external_collections_controller.rb
```diff
@@ -1,7 +1,7 @@
 module Admin
   class ExternalCollectionsController < ApplicationController
     def index
-      @external_collections = Exhibition.all
+      @external_collections = CollectionQuery.new.all_external
```
rfox2: Why not:

```ruby
CollectionQuery.all_external
```

app/queries/collection_query.rb
```diff
@@ -37,6 +37,10 @@ def recent(limit = 5)
     relation.order(updated_at: :desc).limit(limit)
   end
 
+  def all_external
```
rfox2: So, for these methods could we have:

```ruby
def self.all_external
 foo_bar
end
```

app/services/save_collection.rb
```diff
@@ -14,13 +14,7 @@ def save
     fix_image_param!
     collection.attributes = params
     check_unique_id
-
-    if collection.save && process_uploaded_image
```
rfox2: Don't we still need to call process_uploaded_image?

spec/queries/collection_query_spec.rb
```diff
@@ -106,4 +106,12 @@
       subject.build
     end
   end
+
+  describe "all_external" do
+    it "returns all external exhibits" do
+      Collection.new(name_line_1: "test", unique_id: "abc123", url: "http://test").save!
```
rfox2: I'm not strongly opposed to using ActiveRecord for unit tests, but I'd prefer mocking this out to requiring a db connection.

app/controllers/admin/external_collections_controller.rb
jgondron: I was following our existing convention. If we made it a static method then initialize wouldn't get called and relation wouldn't be defined.

app/services/save_collection.rb
jgondron: Yeah this is a merge related issue. I removed this line before those image changes were in. Will fix. Good catch

spec/queries/collection_query_spec.rb
jgondron: I agree with that for most things, except for query testing.

app/controllers/admin/external_collections_controller.rb
rfox2: Not saying we have to do this, I was just wondering why this needed to be an instance method instead of a class method. We could do this:

```ruby
def self.all_external
  relation.where.not(url: nil)
end

private

def relation
 Blah.all
end
```

rfox2:
:+1:

PR #282, opened 2015-10-16
DEC-635 Migrate remaining exhibit methods to collection model

rfox2:
What: Moved remaining exhibit methods to the collection model.
How: Methods were moved and unit tests were updated. Also decided to comment out the exhibit model methods. Other unit tests had to be commented out because they referenced the old methods that are no longer instantiated. Those unit tests will be deleted once we remove the exhibit model from the code.


app/models/exhibit.rb
```diff
@@ -1,25 +1,25 @@
...
+  #
+  # has_paper_trail
+  #
+  # def items_json_url
+  #   "/api/collections/#{collection_id}/items.json?include=image"
+  # end
+  #
+  # def item_json_url(item_id)
+  #   "/api/collections/#{collection_id}/items/#{item_id}.json?include=image"
+  # end
```
jonhartzler: I would say just delete this we have source control.

app/models/exhibition.rb
```diff
@@ -66,17 +66,16 @@ def collection
     @collection ||= get_collection
   end
 
-  # rubocop:disable Metrics/AbcSize
   def save!
     ActiveRecord::Base.transaction do
-      exhibit.collection = collection
+      # exhibit.collection = collection
```
jonhartzler: These as well.

danthewolfe:
:+1:

app/models/exhibit.rb
rfox2: OK

app/models/exhibition.rb
rfox2: Removed

PR #283, opened 2015-11-03
DEC-626: Fix views

jgondron:
All of the data layer changes have been made. These changes are needed to fix the controllers and views to work with those data changes. I also made changes to the behavior of the site setup form. On save, it will now redirect back to the current form for editing. This was taking the user back to the top form, which meant the user would have to click on the form again to see the changes that they made.

### DIFF
--- a/spec/services/save_honeypot_image_spec.rb
+++ b/spec/services/save_honeypot_image_spec.rb
@@ -22,8 +22,8 @@ RSpec.describe SaveHoneypotImage do
   end
 
   let(:honeypot_image) { HoneypotImage.new(item_id: 1) }
-  let(:collection) { double(Collection, id: 1) }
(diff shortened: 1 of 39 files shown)


jonhartzler:
:+1:

spec/services/save_honeypot_image_spec.rb
```diff
@@ -112,4 +117,16 @@
       expect(subject.send(:api_url)).to eq("http://localhost:3019")
     end
   end
+
+  context "group id" do
```
rfox2: What is the group id used for?

rfox2:
:+1:

spec/services/save_honeypot_image_spec.rb
jgondron: Its just a logical grouping of images on the honeypot side. We use collection id as that grouping id.

PR #284, opened 2015-11-04
DEC-627: Remove exhibit and exhibition

jgondron:
Depends on https://github.com/ndlib/honeycomb/pull/283

Now that everything has been changed to not use these objects, I deleted the models, and any leftover references to them within services, views, etc. 

In addition to this cleanup, I found a bug during testing with the url handling in external collections. SaveCollection will now add http:// to an address if it is not given so that links are always treated as absolute.


jonhartzler:
:+1:

config/routes.rb
```diff
@@ -23,8 +22,14 @@
     end
 
     member do
+      site_setup_constraints = { form: /homepage|collection_introduction|about_text|copyright_text/ }
+      get "edit/:form", to: "collections#site_setup", as: :site_setup_form, constraints: site_setup_constraints
+      patch "edit/:form", to: "collections#site_setup_update", as: :site_setup_update_form, constraints: site_setup_constraints
```
rfox2: Is this verb being used correctly? See http://williamdurand.fr/2014/02/14/please-do-not-patch-like-an-idiot/

Should we use PUT instead?

rfox2:
Just had one question about the routes, but apart from that :+1:

PR #285, opened 2015-11-05
DEC-376: Merge collection and exhibit models

jgondron:
Why: Exhibit is no longer needed and is causing extra code complexity, so we are merging it into collection.
How: 
-Migrated about, copyright, enable_browse, enable_search, hide_title_on_home_page, short_description (now named short_intro), description (now named site_intro), url (for external collections), image_*, uploaded_image_*
-Changed Showcases and HoneypotImages to be related to the collection, not the exhibit
-Removed Exhibits from the application layer and changed all service objects, controllers, views to use Collection instead.
-We are intentionally leaving the Exhibits entity at the data layer for now until we are 100% confident that all data was migrated properly in production. There will be a migration later to drop this table.


db/migrate/20151104211754_drop_collection_description.rb
```diff
@@ -0,0 +1,5 @@
+class DropCollectionDescription < ActiveRecord::Migration
```
jgondron: This was accidentally added into the DEC-627 branch. This change is actually associated with DEC-647. We should talk about whether or not we want this to be part of the migration.

app/views/collections/_list.html.erb
```diff
@@ -15,11 +15,11 @@
           <%= CollectionImage.new(collection).display %>
         <td>
           <h4 class="media-heading"><%= h link_to collection.name_line_1, collection_path(collection) %></h4>
-          <%= h collection.description %>
+          <div><%= strip_tags collection.short_intro if collection.short_intro.present? %></div>
```
jgondron: This was accidentally added into the DEC-627 branch. This change is actually associated with DEC-647. We should talk about whether or not we want this to be part of the migration.

jonhartzler:
:+1:

danthewolfe:
:+1:

PR #287, opened 2015-11-05
Settings change for beehive alias

rfox2:
What: Changes settings to that honeycomb points to collections.library.nd.edu as the beehive alias


jgondron:
:+1:

PR #289, opened 2015-11-05
DEC-376: Migration bug fixes

jgondron:
Why: When migrating using data from the production server, I ran into issues with content that contained single and/or double quotes.
How:
-Added a call to quote_string to fix any quoted strings within the existing exhibit data.
-Separated the migrations into adds and copies. Will make it much easier to continue a migration if we find other bugs when deploying to the other servers.


rfox2:
:+1:

jonhartzler:
:+1:

PR #290, opened 2015-11-09
DEC-639: Add pages

jgondron:
Why: We need to add the models and CRUD ops for Pages before we can add to the API/Beehive. 
How: Added a Pages model, controller, routes, views, and all of our typical service class objects such as save, delete, cache key generation, etc. This PR will only give the basics. Extending the API and the redactor to allow for adding images from the collection are part of a different ticket.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -25,12 +25,6 @@ class PagesController < ApplicationController
     end
   end
 
-  def show
-    @page = PageQuery.new.find(params[:id])
-    check_user_edits!(page.collection)
-    redirect_to edit_page_path(page.id)
-  end
-
   def edit
     @page = PageQuery.new.find(params[:id])
     check_user_edits!(@page.collection)
@@ -46,7 +40,7 @@ class PagesController < ApplicationController
 
     if SavePage.call(@page, save_params)
       flash[:notice] = t(".success")
-      redirect_to page_path(@page)
+      redirect_to edit_page_path(@page)
     else
       render :edit
     end
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,0 +1,80 @@
+class PagesController < ApplicationController
+  def index
+    check_user_edits!(collection)
+    @pages = PageQuery.new(collection.pages).ordered
+    cache_key = CacheKeys::Generator.new(key_generator: CacheKeys::Custom::Pages,
+                                         action: "index",
+                                         collection: collection)
+    fresh_when(etag: cache_key.generate)
+  end
+
+  def new
+    check_user_edits!(collection)
+    @page = PageQuery.new(collection.pages).build
+  end
+
+  def create
+    check_user_edits!(collection)
+    @page = PageQuery.new(collection.pages).build(save_params)
+
+    if SavePage.call(@page, save_params)
+      flash[:notice] = t(".success")
+      redirect_to page_path(@page)
+    else
+      render :new
+    end
+  end
+
+  def show
+    @page = PageQuery.new.find(params[:id])
+    check_user_edits!(page.collection)
+    redirect_to edit_page_path(page.id)
+  end
+
+  def edit
+    @page = PageQuery.new.find(params[:id])
+    check_user_edits!(@page.collection)
+    cache_key = CacheKeys::Generator.new(key_generator: CacheKeys::Custom::Pages,
+                                         action: "edit",
+                                         page: @page)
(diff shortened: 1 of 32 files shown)


app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def create # rubocop:disable Metrics/AbcSize
+    check_user_edits!(collection)
+    @page = PageQuery.new(collection.pages).build(save_params)
+
+    if SavePage.call(@page, save_params)
+      flash[:notice] = t(".success")
+      redirect_to page_path(@page)
+    else
+      render :new
```
rfox2: Will the user ever get feedback if the page didn't save? (e.g. - flash[:error] message)

app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  def update
+    @page = PageQuery.new.find(params[:id])
+    check_user_edits!(@page.collection)
+
+    if SavePage.call(@page, save_params)
+      flash[:notice] = t(".success")
+      redirect_to page_path(@page)
+    else
+      render :edit
```
rfox2: Same question here as previous about flash[:error] messages

app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+      redirect_to page_path(@page)
+    else
+      render :new
+    end
+  end
+
+  def show
+    @page = PageQuery.new.find(params[:id])
+    check_user_edits!(page.collection)
+    redirect_to edit_page_path(page.id)
```
rfox2: Why do we have a show action that just redirects to the edit page?

app/services/destroy/page.rb
```diff
@@ -0,0 +1,14 @@
+module Destroy
+  class Page
+    # Destroy the object only
+    def destroy!(page:)
+      page.destroy!
+    end
+
+    # There are no additional cascades for Pages,
+    # so destroys the object only
+    def cascade!(page:)
```
rfox2: Is this a necessary override?

app/views/pages/_list.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    <% pages.each do |page| %>
+      <tr>
+        <td>
+          <h4 class="media-heading"><%= h link_to page.name, page_path(page) %></h4>
+        </td>
+        <td>
+          <%= truncate(strip_tags(page.content), length: 200, omission: '...') %>
+        </th>
+        <td>
+           <%= ListTime.display(page.updated_at) %>
```
rfox2: Whoa, I didn't know we had a ListTime decorator!

spec/services/destroy/page_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    let(:collection) { FactoryGirl.create(:collection) }
+    let(:page) { FactoryGirl.create(:page) }
+
+    # Ensures all data that was created still exists
+    # in the database
+    def data_still_exists!
+      expect { page.reload }.not_to raise_error
+    end
+
+    it "rolls back if an error occurs with Section.destroy!" do
```
rfox2: Do you mean "rolls back if an error occurs with Page.destroy!"?

app/controllers/pages_controller.rb
jgondron: It was to support the route /pages/:id in a similar fashion to showcases. I can remove.

app/services/destroy/page.rb
jgondron: Not sure what you mean. It's more of an expectation that all Destroy objects have a destroy! and a cascade!. In this case, cascade and destroy both do the same thing because there are no other objects to cascade to for a Page, but it still needs to define it.

spec/services/destroy/page_spec.rb
jgondron: Yep, copy paste error

app/controllers/pages_controller.rb
jgondron: The user does get feedback. In the case of errors, the form will highlight the problem fields and give an error message, rather than just a simple flash message.

rfox2: For validation errors, yes. Do we want generalized system errors floating to the top?

app/controllers/pages_controller.rb
rfox2: Could be wrong, but I don't think we need it.

rfox2:
:+1:

jonhartzler:
:+1:

PR #291, opened 2015-11-12
DEC-653: Refactor site ordering

jgondron:
Why: Collection home pages need to be able to show pages, showcases, and items. This can no longer be handled by Showcase order.
How: 
-Added migrations to add the site_objects field to Collections and copy the showcases for the collections into this field as an array of site objects of the format { type, id }
-Created a query object to make it easier to find all site objects for a collection and next/previous object for an object
-Dropped showcase.order
-Added a simple text input for the site_objects in the Site Setup for now until we get an editor created.
-Created a route v1/collections/id/site_objects to get all site objects for a Collection, where the JSON output is similar to v1/collections/id/showcases, but can now be expanded to contain other objects. 
-The previous behavior caused all new showcases to immediately appear on the site. That's no longer the case now that the user has to edit what objects will show on the site. To help with this, I changed the flash messages for page/showcase create to let them know where to go to make the new object show up.

### DIFF
--- a/app/queries/site_objects_query.rb
+++ b/app/queries/site_objects_query.rb
@@ -5,16 +5,26 @@ class SiteObjectsQuery
     end
   end
 
+  # Determines if an object exists in site_objects for its collection
+  def exists?(collection_object:)
+    type = collection_object.class.name
+    id = collection_object.id
+    site_objects_json(collection_id: collection_object.collection_id).detect do |site_object|
+      if site_object[:type] == type && site_object[:id] == id
+        return true
+      end
+    end
+    false
+  end
+
   # Given an object within the collection, this will find the next one in the list
   def next(collection_object:)
-    previous = nil
     type = collection_object.class.name
     id = collection_object.id
     site_objects_json(collection_id: collection_object.collection_id).each_cons(2) do |site_object, next_site_object|
       if site_object[:type] == type && site_object[:id] == id
         return get_object(site_object: next_site_object)
       end
-      previous = site_object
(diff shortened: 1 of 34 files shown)


jonhartzler:
:+1:

app/queries/site_objects_query.rb
```diff
@@ -0,0 +1,62 @@
+class SiteObjectsQuery
+  def all(collection:)
+    site_objects_json(collection_id: collection.id).map do |site_object|
+      get_object(site_object: site_object)
+    end
+  end
+
+  # Given an object within the collection, this will find the next one in the list
+  def next(collection_object:)
+    previous = nil
```
rfox2: Same comment from the other PR. Is previous necessary in the context of this method?

app/views/collections/_homepage_form.html.erb
```diff
@@ -1,6 +1,7 @@
 <%= display_errors(@collection) %>
 
 <%= f.input :short_intro, input_html: { class: 'honeycomb_redactor'  } %>
+<%= f.input :site_objects, as: :text %>
```
rfox2: I get it now. :+1:

rfox2:
:+1:  There's a lot of duplication between this and PR 292. Once the questions are answered from that one, you could merge both of these in.

app/queries/site_objects_query.rb
jgondron: I don't think it is, I'll remove.

PR #292, opened 2015-11-13
DEC-652: Build pages api

jgondron:
Why: Now that page models have been created, need a way to get page data via the api
How: 
-Added routes/controllers/views for v1 pages index and show
-Added cachekey generators for v1 pages index/show
-Changed showcases#show to use site objects for getting next/previous objects
-Added additionalType to item and section for consistency

Depends on https://github.com/ndlib/honeycomb/pull/291, will rebase when that's merged in.

### DIFF
--- a/spec/decorators/v1/page_json_decorator_spec.rb
+++ b/spec/decorators/v1/page_json_decorator_spec.rb
@@ -1,0 +1,68 @@
+require "rails_helper"
+
+RSpec.describe V1::PageJSONDecorator do
+  subject { described_class.new(page) }
+
+  let(:page) { double(Page) }
+
+  describe "generic fields" do
+    [:id, :unique_id, :collection, :updated_at, :name, :content].each do |field|
+      it "responds to #{field}" do
+        expect(subject).to respond_to(field)
+      end
+    end
+  end
+
+  describe "#at_id" do
+    let(:page) { double(Page, unique_id: "page_id") }
+
+    it "returns the path to the id" do
+      expect(subject.at_id).to eq("http://test.host/v1/pages/page_id")
+    end
+  end
+
+  describe "#collection_url" do
+    let(:collection) { double(Collection, unique_id: "collection_id") }
+    let(:page) { double(Page, collection: collection) }
+
+    it "returns the path to the items" do
+      expect(subject.collection_url).to eq("http://test.host/v1/collections/collection_id")
+    end
+  end
+
+  describe "#slug" do
+    let(:page) { double(Page, slug: "sluggish") }
+
+    it "calls the slug generator" do
+      expect(CreateURLSlug).to receive(:call).with(page.slug)
+      subject.slug
+    end
+  end
+
+  describe "#display" do
+    let(:json) { double }
+
+    it "calls the partial for the display" do
+      expect(json).to receive(:partial!).with("/v1/pages/page", page_object: page)
+      subject.display(json)
+    end
+  end
+
+  describe "#next" do
+    it "uses the site objects query to retrieve the next object" do
+      expect_any_instance_of(SiteObjectsQuery).to receive(:next).with(collection_object: page)
+      subject.next
+    end
+  end
+
+  describe "#previous" do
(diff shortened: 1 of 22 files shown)


app/queries/site_objects_query.rb
```diff
@@ -0,0 +1,62 @@
...
+  # Given an object within the collection, this will find the next one in the list
+  def next(collection_object:)
+    previous = nil
+    type = collection_object.class.name
+    id = collection_object.id
+    site_objects_json(collection_id: collection_object.collection_id).each_cons(2) do |site_object, next_site_object|
+      if site_object[:type] == type && site_object[:id] == id
+        return get_object(site_object: next_site_object)
+      end
+      previous = site_object
```
rfox2: Is this necessary? The method doesn't reference previous anywhere, but nil will get returned if the iteration fails. Does previous need to be in this method at all?

app/views/collections/_homepage_form.html.erb
```diff
@@ -1,6 +1,7 @@
 <%= display_errors(@collection) %>
 
 <%= f.input :short_intro, input_html: { class: 'honeycomb_redactor'  } %>
+<%= f.input :site_objects, as: :text %>
```
rfox2: What does this form element do?

spec/decorators/v1/page_json_decorator_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    it "returns the path to the items" do
+      expect(subject.collection_url).to eq("http://test.host/v1/collections/collection_id")
+    end
+  end
+
+  describe "#slug" do
+    let(:page) { double(Page, slug: "sluggish") }
+
+    it "calls the slug generator" do
+      expect(CreateURLSlug).to receive(:call).with(page.slug)
```
rfox2: Is the slug a derivative of something? Just curious. If it is, should we test this to determine whether the correct slug is returned?

spec/decorators/v1/page_json_decorator_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  describe "#display" do
+    let(:json) { double }
+
+    it "calls the partial for the display" do
+      expect(json).to receive(:partial!).with("/v1/pages/page", page_object: page)
+      subject.display(json)
+    end
+  end
+
+  describe "#next" do
```
rfox2: We should test for the null scenario for this and #previous.

spec/models/page_spec.rb
```diff
@@ -18,4 +18,9 @@
     expect(subject).to respond_to(:paper_trail_enabled_for_model?)
     expect(subject.paper_trail_enabled_for_model?).to be(true)
   end
+
+  it "uses name for the slug" do
```
rfox2: So the slug is based on the name? Could we have a similar test in the page json decorator?

spec/decorators/v1/page_json_decorator_spec.rb
jgondron: Yeah, our current pattern requires models to have a slug method. When calling CreateURLSlug we pass this along with it, so at this point in the code all we care about is that the json decorator uses CreateURLSlug with the model's slug. As far as the json decorator is concerned, it doesn't care what attribute is actually used to determine the slug.

spec/models/page_spec.rb
jgondron: See other comment in the json decorator.

spec/decorators/v1/page_json_decorator_spec.rb
jgondron: Do you mean testing for the case when the return value is nil? Or the case when the passed in object is nil? With regards to return values, I'm testing that in the SiteObjectsQuery spec. For the passed in object, I'm not sure why any caller would ever try to decorate a nil object.

jgondron: Also note that if next is nil, jbuilder seems to exclude it from the rendered json. I tested that by hitting the v1 routes and examining the output (since we aren't testing render output in our specs)

PR #293, opened 2015-11-19
DEC-654: Add representational images to Pages

jgondron:
Why: Need to add background/representational image to a Page. This image will be used when rendering a Page cards such as on the homepage or "next" links.
How: We decided to start migrating toward having an Image object that can be used for Items, and background images for Collections, Showcases, Pages, etc. 
-Created an Image model and an association to Pages. 
-Changed SavePage to find or create an Image using the uploaded image and associate it to the page. If the image was previously uploaded, it will reuse the image, otherwise it will create a new one.
-Started with the basic styles needed, we may need to expand this later as we migrate the other objects, such as for dzi files.


jonhartzler:
:+1:

rfox2:
:+1:

PR #294, opened 2015-11-19
Dec 650 images on site pages

rfox2:
What: Images can now be added to site pages / essays either by selecting an image that was previously uploaded and made into an item, or by selecting an image from the local system which will then be converted into a collection item and inserted into the redactor document.

* a plugin for redactor was created to allow selection of existing image items
* action was added to the item controller that sends out a json list of items associated with a collection
* action was added to the collection controller to upload a new image and convert it into an item
* redactor was modified to include the image button on the page edit interface and the intro edit interface
* also had to update the package.json file to accommodate for version compatibility


app/controllers/v1/items_controller.rb
```diff
@@ -12,6 +12,19 @@ def index
       fresh_when(etag: cache_key.generate)
     end
 
+    def images
```
rfox2: I've broken this up

app/controllers/collections_controller.rb
```diff
@@ -108,8 +108,38 @@ def unpublish
     end
   end
 
+  def image_upload
```
rfox2: I've broken this up

app/controllers/collections_controller.rb
```diff
@@ -108,8 +108,24 @@ def unpublish
...
```
jgondron: Any reason we couldn't use the items controller to create the item?

rfox2: I used the collections controller because it already had the collection unique_id in the route. I could be persuaded otherwise.

jonhartzler: the route to create items is 

/collections/id/items  so the id should be in the route.   

not to be confused with the show item routes which are /items/id   

So I think we should try and get the new in the route.

app/controllers/v1/items_controller.rb
```diff
@@ -12,6 +12,12 @@ def index
...
```
jgondron: v1/collections/:id/items already gives the image data within the items array. Could probably have reused that instead of creating this new method.

app/controllers/v1/items_controller.rb
```diff
@@ -45,6 +59,10 @@ def showcases
 
     protected
 
+    def save_item_params
```
jgondron: Can you use save_params?

app/controllers/v1/items_controller.rb
```diff
@@ -21,6 +21,20 @@ def show
...
     end
 
+    def create
+      @collection = CollectionQuery.new.any_find(params[:collection_id])
+      @item = ItemQuery.new(@collection.items).build
+
+      return if rendered_forbidden?(@item.collection)
+
+      if SaveItem.call(@item, save_item_params)
+        flash[:success] = "Item created"
```
jgondron: Not sure if we should use flash in the api

app/controllers/v1/items_controller.rb
```diff
@@ -21,6 +21,20 @@ def show
...
+        render :create
+      else
+        render json: { status: "error" }, status: 500
```
jgondron: We should probably render errors similar to the way update works. Update renders the item json back to the client and adds an array of errors associated with the item.

app/controllers/collections_controller.rb
```diff
@@ -110,6 +110,10 @@ def unpublish
 
   protected
 
+  def save_item_params
```
jgondron: Can we remove this now that you're able to use save_params?

config/routes.rb
```diff
@@ -25,6 +25,7 @@
       site_setup_constraints = { form: /homepage|collection_introduction|about_text|copyright_text/ }
       get "edit/:form", to: "collections#site_setup", as: :site_setup_form, constraints: site_setup_constraints
       put "edit/:form", to: "collections#site_setup_update", as: :site_setup_update_form, constraints: site_setup_constraints
+      post :image_upload
```
jgondron: Can this be removed now?

app/controllers/collections_controller.rb
rfox2: Done

config/routes.rb
rfox2: Done

jgondron:
:+1:

PR #295, opened 2015-11-30
DEC-684: Unable to save metadata

jgondron:
Problem was due to the items controller not allowing arrays for some of the metadata fields that allow multiples. Fixed the items controller to allow arrays as params for all multiple fields. Also changed to permit user_defined_id to allow the user to change this if necessary.


jonhartzler:
:+1:

rfox2:
:+1:

PR #296, opened 2015-11-30
DEC-685: Showcase order default

jgondron:
ShowcaseQuery will now add a default value for order that is based on the existing showcase orders when calling the build method. This should effectively append any new showcases to the end of the exhibit's list of showcases. Since it's done in the build method, the user has the ability to change it before saving the new showcase.


jonhartzler:
:+1:

rfox2:
:+1:

PR #299, opened 2015-12-09
Dec 682 associate pages and items

rfox2:
What: Added functionality to associate pages and items (images)
How:

* Created join model
* Created services to destroy associations
* Created services to manage page parsing on saves
* Added code that checks for page associations when an item image is updates, and replaces the image reference on associated pages
* Added a UI panel to indicate when an item is associated with one or more pages, and disables the delete button for that item
* Refactored V1 API output to include associated items for pages


app/services/parse_page.rb
```diff
@@ -0,0 +1,13 @@
+class ParsePage
```
jonhartzler: Consider removing this class or making it a private class of ReplacePageItem since it only returns a nokogiri object, is bound to the content delivered by, and is stored as a private member of that class.

rfox2: It's currently used only by that class but could be used in future classes where we need a parsed page for other purposes. I lean toward making it a private class if want to go that route. Is your preference for elimination?

jonhartzler: Private works for me.  It can be made public if a need arrises.

app/assets/javascripts/components/PagesPanel.jsx
```diff
@@ -0,0 +1,83 @@
...
+    if(page.image) {
+      return (
+        <div className="image">
+          <Avatar src={ page.image["thumbnail/small"]["contentUrl"] } />
+        </div>
+      )
+    } else {
+      return (
+        <div className="image" >
+          <Avatar>{ "P" }</Avatar>
```
jgondron: On showcases, we used the showcase name as the letter to display. Should we use page name? If not, should we change ShowcasePanel to use 'S'?

jgondron:
Should also update en.yml, items.delete_panel.cannot_delete_message to something more appropriate now that an items association to a page can also prevent a delete.

jgondron:
Also need to update CacheKeys Items.edit to include the association. Otherwise, if you add an item to a page, load item edit, delete the page or item from the page, then reloading the item edit will still show the old association.

app/services/destroy/item.rb
```diff
@@ -19,6 +19,9 @@ def cascade!(item:)
         item.sections.each do |child|
           @destroy_section.cascade!(section: child)
         end
+        item.pages.each do |page|
```
jgondron: Should use a DestroyPage object and call cascade. This is leaving orphaned page associations.

app/services/destroy/item.rb
```diff
@@ -19,6 +19,9 @@ def cascade!(item:)
...
+          item.pages.delete(page)
```
jgondron: Should use a DestroyPage object and call cascade. This is leaving orphaned page associations.

jgondron: I also don't think this should delete the page actually. Shouldn't it just remove the association and strip the item out of the page? The analog would be deleting Sections inside a Showcase, instead of the entire Showcase.

app/services/destroy_page_item_associations.rb
```diff
@@ -0,0 +1,17 @@
...
+  def self.call(page_id:)
+    new(page_id: page_id).delete!
+  end
+
+  def initialize(page_id:)
+    @page = Page.find(page_id)
+  end
+
+  def delete!
+    @page.items.delete_all
```
jgondron: Shouldn't this be deleting items_pages, not items?

db/migrate/20151125174835_create_join_table_pages_items.rb
```diff
@@ -0,0 +1,10 @@
+class CreateJoinTablePagesItems < ActiveRecord::Migration
+  def change
+    create_join_table :pages, :items do |t|
+      t.integer :item_id
+      t.integer :page_id
+      t.index [:page_id, :item_id]
+      t.index [:item_id, :page_id]
+    end
+  end
```
jgondron: Add an FKC between items_pages and items. We don't want to be able to delete items if an association exists, regardless of our checks at the app layer.

spec/controllers/pages_controller_spec.rb
```diff
@@ -209,13 +213,13 @@
...
 
-    it "uses page query " do
+    it "uses page query" do
       expect_any_instance_of(PageQuery).to receive(:find).with("1").and_return(page)
       subject
     end
 
     it "uses the Destroy::Page.cascade method" do
-      expect_any_instance_of(Destroy::Page).to receive(:cascade!)
+      # implicit call to cascade! triggering a call to DestroyPageItemAssociations
```
jgondron: It should still use a DestroyPage service object to destroy the page.

spec/controllers/pages_controller_spec.rb
```diff
@@ -189,6 +189,10 @@
   describe "DELETE #destroy" do
     subject { delete :destroy, id: page.id }
 
+    before(:each) do
+      expect(DestroyPageItemAssociations).to receive(:call).and_return(1)
```
jgondron: Should this be an expect or an allow? Seems like we should have an expectation that Destroy::Page is used here. If we expect Destroy::Page to use DestroyPageItemAssociations, that expectation would be in the spec for Destroy::Page, not here.

jgondron:
Missing specs for these files that changed:
decorators/v1/page_json_decorator.rb
models/item.rb
models/items_pages
models/page

Also need specs in all of the destroy methods to test that it ensures integrity. Ex: DestroyItem throws exception if an association exists, DestroyPage throws exception if an association exists. See the "cascade transaction" contexts within these specs.

app/assets/javascripts/components/PagesPanel.jsx
rfox2: The problem with showing the name is that longer names (meaning > 4 char) are almost illegible, thus meaningless. If we're going to have a placeholder, I'd rather switch to a single letter.

jgondron: Just to clarify, we used the first letter of the name, not the full name.

rfox2: I changed it so that it uses the first letter of the page name. Sorry about that.

app/services/destroy/item.rb
rfox2: This actually just removes the association, not the page itself.

spec/controllers/pages_controller_spec.rb
rfox2: Changed it to be allow instead of expect

db/migrate/20151125174835_create_join_table_pages_items.rb
rfox2: Added the dual FKCs

spec/controllers/pages_controller_spec.rb
rfox2: The spec was corrected to include this again

jonhartzler:
:+1:

PR #301, opened 2015-12-21
Fix api url

jgondron:
API is currently using collections.library.nd.edu for things like showcases/items parts of collections, should be using honeycomb.library.nd.edu

Note: Previous PR was to fix 2.0


jonhartzler:
:+1:

PR #302, opened 2016-01-18
DEC-674: Migrate developers to postgres

jgondron:
Why: We are going to be working on some features such as dynamic metadata that will be easier to r/w with postgres
How:
-Added pg gem to support postgres
-Fixed migrations and queries that explicitly used sql since there are differences between mysql/pg
-Fixed text type definitions, the pg equivalent of mysql longtext is just text
-Fixed travis to work with pg
-Found a problem with the section query and changed to work a little more safely with both pg and mysql
-Found a problem with ordering differences between postgres and mysql. Changed the default order of collections to force it to match the way mysql collections were ordered.

Specific to server migration:
-Created a pgloader script to use to copy the data. Had to make it only copy data to avoid it writing the schema differently from rails. This means we'll need to use rails to create the db and load the schema, then run the pgloader script to handle data copy.
-Wrote a basic scraper in python to get api data and wrote a comparator to compare the json files. These will be used to perform a pre/post server migration comparison to make sure the output is the same.
-For server migration, will need to drop constraints, copy data, then recreate the constraints. Added a script that will generate a drop/create script that we can use during the migration.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem "responders", "~> 2.0"
 # gem "therubyracer",  platforms: :ruby
 
 gem "mysql2"
+gem "pg"
 
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 gem "jbuilder", "~> 2.0"
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,6 @@ gem "responders", "~> 2.0"
 # See https://github.com/sstephenson/execjs#readme for more supported runtimes
 # gem "therubyracer",  platforms: :ruby
 
-gem "mysql2"
 gem "pg"
 
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
--- a/app/queries/section_query.rb
+++ b/app/queries/section_query.rb
@@ -20,10 +20,12 @@ class SectionQuery
   end
 
   def next(section)
-    relation.where(showcase_id: section.showcase_id).where("`#{relation.table_name}`.order > ?", section.order).order(:order).first
+    order_field = Section.arel_table[:order]
+    relation.where(showcase_id: section.showcase_id).where(order_field.gt(section.order)).order(order: :asc).first
   end
 
   def previous(section)
-    relation.where(showcase_id: section.showcase_id).where("`#{relation.table_name}`.order < ?", section.order).order(order: :desc).first
+    order_field = Section.arel_table[:order]
+    relation.where(showcase_id: section.showcase_id).where(order_field.lt(section.order)).order(order: :desc).first
   end
 end
--- a/db/dec_json_comparator.rb
+++ b/db/dec_json_comparator.rb
@@ -1,0 +1,39 @@
+# Compares two json files for differences using HashDiff. Order of arrays does not matter.
+# No output will be given if no differences were found.
+require 'json'
+require 'hashdiff'
+
+# Deep sorts jsons by id's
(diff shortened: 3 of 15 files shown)


db/dec_json_comparator.rb
```diff
@@ -0,0 +1,39 @@
...
+sort!(json1)
+file2 = File.read(file2_path)
+json2 = JSON.parse(file2)
+sort!(json2)
+
+diff = HashDiff.diff(json1, json2)
+if diff.any?
+  print "HashDiff of '#{file1_path}' and '#{file2_path}':\n"
+  print HashDiff.diff(json1, json2)
+  print "\n"
```
jgondron: This isn't run within rails, ignore this.

Gemfile
```diff
@@ -12,6 +12,7 @@ gem "responders", "~> 2.0"
 # gem "therubyracer",  platforms: :ruby
 
 gem "mysql2"
```
jonhartzler: remove this gem ?

jonhartzler:
:+1:

app/queries/section_query.rb
```diff
@@ -20,10 +20,12 @@ def build(args = {})
...
 
   def next(section)
-    relation.where(showcase_id: section.showcase_id).where("`#{relation.table_name}`.order > ?", section.order).order(:order).first
+    order_field = Section.arel_table[:order]
+    relation.where(showcase_id: section.showcase_id).where(order_field.gt(section.order)).order(order: :asc).first
   end
 
   def previous(section)
-    relation.where(showcase_id: section.showcase_id).where("`#{relation.table_name}`.order < ?", section.order).order(order: :desc).first
+    order_field = Section.arel_table[:order]
```
rfox2: Is arel_table part of a PG driver associative array?

db/migrate/20150604150434_add_metadata_to_items.rb
```diff
@@ -1,5 +1,10 @@
 class AddMetadataToItems < ActiveRecord::Migration
   def change
-    add_column :items, :metadata, :text, limit: 4294967295
+    case ActiveRecord::Base.connection.class.name
+    when /.*Mysql.*/
```
rfox2: Does this just check to see if the driver being used is mysql?

db/pgloader.lisp
```diff
@@ -0,0 +1,6 @@
+LOAD DATABASE
```
rfox2: What is this used for?

db/dec_scraper.py
```diff
@@ -0,0 +1,46 @@
+import urllib
```
rfox2: What is this used for?

rfox2:
:+1:

db/dec_scraper.py
jgondron: http lib for making the requests to the api

db/migrate/20150604150434_add_metadata_to_items.rb
jgondron: yes

app/queries/section_query.rb
jgondron: Library used by rails to generate queries. Here I'm using it to generate the sql for Section.order since there are differences in quoting identifiers in mysql vs pg.

jonhartzler: No arel is a part of rails.  It is a DSL for sql.  

https://github.com/rails/arel

PR #303, opened 2016-01-19
DEC-695: Cannot insert items into page if collection is not published or preview enabled

jgondron:
Why: The page redactor allows the user to insert images that are in the collection, but this fails if the collection is not either in preview mode or is published.
How: Modified ItemsController#index to use any_find instead of public_find, so that we can get items for the collection even if it's not published/preview enabled.


jonhartzler:
:+1:

rfox2:
:+1:

PR #304, opened 2016-01-20
DEC-696: Page/Showcase edit fails to load when site objects is nil

jgondron:
Why: Within Page/Showcase edit, the delete button tries to disable itself if the object is included in site_objects, as a way of enforcing referencial integrity at the app layer. This check fails when the field is nil.
How: Changed the site objects query object to return an empty array if the field is null or empty string.


jonhartzler:
:+1:

rfox2:
:+1:

PR #305, opened 2016-01-20
Dec 731 fix display for items without images

jonhartzler:
Why: this is necessary so we can display correctly in the UI when an Item does not have an image associated with it.  This is part of the transition to having items that do not have images.

What: I added a new status to the image status. Changed where that status gets set in save item. Updated the listing page and item edit form to look at the status.  Re added a timeout after 7 tries to display an error message prompting the user to take action about an image that may be in error.  I also added a check to test if there is no sneakers connection and put the image in an error state.   

What i did not do: 
Create any way to add items with out images other than the import
Create any way for the user to delete an existing image therefore putting it in a state where it has no image.


app/assets/javascripts/components/ItemShowImageBox.jsx
```diff
@@ -72,55 +42,77 @@ var ItemShowImageBox = React.createClass({
       dataType: "json",
       method: "GET",
       success: (function(data) {
-        this.checkImageState(data);
-        this.setState({ awaitingResponse: false });
+        if (this.state.requestCount < this.props.maxRetries) {
+          this.setState({ item: data.items, awaitingResponse: false }, this.testImageStatus);
+        } else {
+          var item = this.state.item;
+          item.image_status = "image_invalid";
```
rfox2: Just making sure I'm reading this correctly. After 7 tries, the image status will be marked as "image_invalid"? Are we assuming that the problem is with the image or could it be with the service itself?

app/assets/javascripts/components/ItemShowImageBox.jsx
```diff
@@ -72,55 +42,77 @@ var ItemShowImageBox = React.createClass({
...
-    if(this.props.item.image_status == "image_ready") {
-      this.setState({
-        imageReady: true,
-      });
-    } else if(this.props.item.image_status == "image_invalid") {
+    this.testImageStatus();
+  },
+
+  testImageStatus: function () {
+    if (this.state.item.image_status == "image_invalid") {
```
rfox2: Oh, I see. You're just using it as a flag for a conditional...

app/services/save_item.rb
```diff
@@ -56,7 +56,14 @@ def fix_image_param!
   def process_uploaded_image
     if params[:uploaded_image]
       item.image_processing!
-      QueueJob.call(ProcessImageJob, object: item)
+      begin
+        QueueJob.call(ProcessImageJob, object: item)
+      rescue Bunny::TCPConnectionFailedForAllHosts
+        item.image_invalid!
```
rfox2: Again, just making sure I understand. So when there is a network error we're going to mark the image as invalid?

rfox2:
:+1:

jgondron:
:+1:

PR #306, opened 2016-01-21
Dec 738 migrate metadata to jsonb

jonhartzler:
why: now that the we have jsonb. we need to migration of the old json store of metadata values to the new and flashy jsonb fields. 

what: added a migration to convert the field and converted the class.


rfox2:
:+1:

jgondron:
:+1:

PR #307, opened 2016-01-22
Dec 739 create collection metadata config

jonhartzler:
Why:  Config needs to be stored in the database.  

What:  Updated to auto generate the configuration when a collection is created and to store it in the database.


jgondron:
:+1:

rfox2:
:+1:

PR #311, opened 2016-02-01
Fix travis build

jgondron:
Why: Build is currently broken because of fsevents dependency. 
How: fsevents is a dep of gulp-watch. We don't need gulp in honeycomb. Removed from package.json and regenerated the shrinkwrap.


danthewolfe:
:+1:

jonhartzler:
:+1:

PR #312, opened 2016-02-01
DEC-773: Add metadata drop down doesn't appear

jgondron:
Why: Drop down box for adding new metadata has disappeared from the item form.
How: Readded the ItemMetadataSelectAdditionalFields to the metadata form. Looks like it may have been removed by accident during the mui update.


jonhartzler:
:+1:

rfox2:
:+1:

PR #308, opened 2016-01-27
DEC-655: Site path interface

jgondron:
Why: The user needs an easy way to manage the site path for a collection.
How: Created a drag and drop interface that allows the user to drag any showcase or page into the site path. 

Note: Several files were changed that are seemingly unrelated to this feature because of react and mui updates.

PR #313, opened 2016-02-01
DEC-759: User needs to view current metadata configuration for a collection

jgondron:
Note: This is based on https://github.com/ndlib/honeycomb/pull/308. Review/merge that first.

Why: Now that each collection can customize their metadata, we need to start off with giving a view of the current configuration.
How: Created a settings form/subform system similar to site setup so that we can break up the collection settings into multiple views. Created a basic table view of the current metadata configuration.


jonhartzler:
:+1:

rfox2:
:+1:

PR #298, opened 2015-12-04
BugFix: Search and Browse Stretched Images

jgondron:
Why: Related to https://docs.google.com/document/d/1yzEkvolhxhzUDvHD4y92hOvj-g4yx5lcgO7MeFJu0wU/edit#bookmark=id.8t3ocag3vsri
How: Beehive requires the medium url for search hits. Changed the waggle item. This change will require reindexing all items so that the new url is returned in search results.

Note: This PR is for a bug fix on 2.0. Ticket https://jira.library.nd.edu/browse/DEC-688 should address the problem in master.

PR #314, opened 2016-02-02
DEC-774: Bugfix stretched images in search/browse

jgondron:
Similar to https://github.com/ndlib/honeycomb/pull/298, this is to patch master.

Note again, this will require reindexing all of your items so that the medium url is returned in search results (rake search:index_all).


jonhartzler:
:+1:

rfox2:
:+1:

PR #315, opened 2016-02-03
Dec 756 map config to api

jonhartzler:
Refactor to make the metadata configuration configurable per collection.

Basic What:

1.  Changed all out going references to configuration to ask the collection what its config is.  
2.  Made saving and updating work again.  By changing name to be a true metadata field.   
3.  Changed a lot of tests to reference doubles rather than the items.yml config.
4.  Changed the app to use the items.yml to setup a default config value. 

Still to do 
Remove public interface to metadata in item.
Make Description act like name.
Move configuration from its own database table to a part of collection. 
Move user_defined_id out of metadata.


app/controllers/v1/collections_controller.rb
```diff
@@ -55,7 +55,9 @@ def preview_mode
     end
 
     def metadata_configuration
-      @configuration = Metadata::Configuration.item_configuration
+      collection = CollectionQuery.new.any_find(params[:collection_id])
+      @configuration = CollectionConfigurationQuery.new(collection).find
+      @configuration = Metadata::Configuration.new(@configuration)
```
rfox2: what happens during the successive mutation?

app/models/item.rb
```diff
@@ -34,8 +36,8 @@ class Item < ActiveRecord::Base
   has_attached_file :uploaded_image,
                     restricted_characters: /[&$+,\/:;=?@<>\[\]{}\|\\^~%#]/
 
-  validates :name, :collection, :unique_id, :user_defined_id, presence: true
-  validates :date_created, :date_modified, :date_published, date: true
+  validates :collection, :unique_id, :user_defined_id, presence: true
+  # validates :date_created, :date_modified, :date_published, date: true
```
rfox2: can this be deleted?

app/models/item.rb
```diff
@@ -43,6 +45,14 @@ class Item < ActiveRecord::Base
 
   enum image_status: { image_invalid: 0, image_processing: 1, image_ready: 2 }
```
rfox2: unrelated to pull request, but are we still using "image_invalid" as a descriptor?

app/controllers/v1/collections_controller.rb
jonhartzler: This is a non dry usage that I have corrected in the next Pull Request.

app/models/item.rb
jonhartzler: YES !!

app/models/item.rb
jonhartzler: no it is no_image  that is in the pull request i merged in the morning.

app/models/metadata/retrieval.rb
```diff
@@ -0,0 +1,87 @@
...
+    def metadata_value(field)
+      value = get(field)
+      config = configuration.field(field)
+
+      return nil if value.nil? || config.nil?
+
+      type = config.type
+      value = ensure_value_is_array(value)
+
+      if type.to_sym == :string
```
rfox2: could this conditional sequence become a private method?

app/models/waggle.rb
```diff
@@ -24,6 +24,19 @@ def self.commit
   end
 
   def self.search(**args)
+    set_configuration(::Metadata::Configuration.new(CollectionConfigurationQuery.new(args[:collection]).find))
```
rfox2: noob question: why does Metadata::Configuration need to be namespaced this way?

app/services/create_items.rb
```diff
@@ -28,7 +28,7 @@ def create_or_update!(collection_id:, find_by:, items_hash:, counts:, errors:)
     ActiveRecord::Base.transaction do
       items_hash.each.with_index do |item_props, index|
         rewrite_errors = []
-        item_props = yield(item_props, rewrite_errors) if block_given?
+        item_props = yield(item_props, rewrite_errors).symbolize_keys if block_given?
```
rfox2: when you get a chance could you explain how this expression works?

app/models/waggle.rb
jonhartzler: I did it to follow what was already started with Jaron's work.   He created a metadata section.   I have been trying to use that more.  Also, Metadata indicates it is in a directory.  So to create a directory I have to do this namespacing.

app/models/metadata/retrieval.rb
jonhartzler: I thought about that I will do it.

app/services/metadata_input_cleaner.rb
```diff
@@ -0,0 +1,19 @@
+class MetadataInputCleaner
```
rfox2: not to be picky - this name seems a little confusion. this service class is essentially decomposing an array value into the item metadata values using a hash.  how about MetadataInputDecomposition - it's a mouthful, i know

spec/models/waggle/adapters/solr/search/result_spec.rb
```diff
@@ -109,8 +115,7 @@
 
       it "is the configured facet fields, excluding their relevant filter" do
         expect(subject).to eq([
-          "{!ex=creator_facet}creator_facet",
-          "{!ex=language_facet}language_facet"
+          "{!ex=creator_facet}creator_facet"
```
rfox2: I think rubocop just wants the left and right brackets to be on their own line with the values indented inside of the array brackets

rfox2:
I just had one remaining question about the naming of MetadataInputCleaner, but that's not a show stopper. All other questions are answered! :-)

Once the hound issues are looked at I'd say :+1:

app/models/item.rb
```diff
@@ -34,8 +36,7 @@ class Item < ActiveRecord::Base
   has_attached_file :uploaded_image,
                     restricted_characters: /[&$+,\/:;=?@<>\[\]{}\|\\^~%#]/
 
-  validates :name, :collection, :unique_id, :user_defined_id, presence: true
```
jgondron: Is something else now handling validation of these fields? Even if we're moving them to metadata, if the expectation was that the app was validating their presence, then we should continue to do so.

jonhartzler: Yes.  The metadata configuration manages this.  All things metadata need to be managed in metadata not duplicated into places that no longer have or manage that field.   Name is no longer considered a part of item.  The only reason that it is mapped in  is for some of the admin views that display name and description (think the item list).   It would be fixed by using the api completely.

app/services/create_items.rb
jgondron: Wondering if the injected block shouldn't handle symbolizing the keys. The purpose of these injected blocks is to transform the item hash into something we can directly pass to an item save, so it seems reasonable to expect it to symbolize the keys.

spec/controllers/v1/collections_controller_spec.rb
```diff
@@ -210,10 +237,5 @@
       expect(SaveCollection).to receive(:call).with(collection, site_objects: site_objects_translated).and_return(true)
       subject
     end
-
-    it "checks the editor permissions" do
```
jgondron: Why was this removed?

spec/decorators/v1/item_json_decorator_spec.rb
```diff
@@ -26,7 +26,7 @@
   end
 
   describe "#at_id" do
-    let(:item) { double(Item, unique_id: "adsf") }
+    let(:item) { double(Item, unique_id: "adsf", name: "name") }
```
jgondron: I'm still a little concerned with using double instead of instance_double. When we remove the store_accessors from Item these specs will still pass, which doesn't give us very good regression testing.

spec/validators/date_validator_spec.rb
```diff
@@ -1,5 +1,8 @@
 require "rails_helper"
 
+# commented out because it is not in use.  Not removed because I think I want to bring it
+# back in the next refactor
+=begin
```
jgondron: Do not comment about usage of block comments.

spec/controllers/v1/collections_controller_spec.rb
jonhartzler: Because metadata configuration is not an editor only action.

spec/decorators/v1/item_json_decorator_spec.rb
jonhartzler: updated

spec/validators/date_validator_spec.rb
jonhartzler: I just wanted it to be clear why I did not just remove this file.   Because the whole file is commented out.

PR #316, opened 2016-02-05
DEC-760: Edit metaconfig form

jgondron:
Why: User needs a way to modify existing metadata configuration fields.
How: When viewing metadata settings, a user can now click on an edit button that will bring up a dialog window that allows them to modify the settings for that field. It is currently all wired up to call an action that changes the store and submits an ajax request to update the field. There is no route to this update point in the api at the moment, so all requests will fail. This will need to be updated during work on DEC-762. An additional note: The appearance of this form and the table view will need some work once we get our themes fixed (DEC-761).


app/assets/javascripts/components.js
```diff
@@ -5,6 +5,7 @@ React = require("react");
 ReactDOM = require("react-dom");
 
 // mixins
+LinkedStateMixin = require('react-addons-linked-state-mixin');
```
rfox2: Could we try using higher-order components instead of mixins? See https://medium.com/@dan_abramov/mixins-are-dead-long-live-higher-order-components-94a0d2f9e750#.33dvmjh0s

app/assets/javascripts/constants/MetaDataConfigurationActionTypes.js
```diff
@@ -0,0 +1,6 @@
+var keyMirror = require('keymirror');
```
rfox2: How does this fit in the bigger picture Justin? A quick overview would be helpful.

rfox2:
Just some questions and comments. I love the use of shrinkwrap. I'm trying to get a sense of how these pieces fit together, but the code looks fine as far as I can tell. I would like to understand how MetaDataConfigurationActionTypes.js fits in overall - is it a couple of constants? After all of my questions have been thoroughly and completely answered in dissertation format, then :+1:

rfox2:
Oh, and the hound issues are fixed, then :+1:

app/assets/javascripts/components.js
jgondron: I didn't end up using the actual mixin they provide, since I was doing something a little more complex than binding it to state properties. I left this in in case we wanted to use it in other places, but I'll take it out for now. For more info on this, https://facebook.github.io/react/docs/two-way-binding-helpers.html. It's pretty useful when creating forms with react components.

app/assets/javascripts/constants/MetaDataConfigurationActionTypes.js
jgondron: The gist is that an action is a request to change a store in some way. This is done by sending messages through the app dispatcher. This means multiple things need to know about the action types, at the very least the action class itself, and the store. This class just defines the types of messages so that all of those objects can use the same values. Otherwise we have magic numbers/strings everywhere. 

I can review this more when in the office, but this is a part of flux so it might be helpful to review the diagrams here https://facebook.github.io/flux/docs/overview.html.

app/assets/javascripts/actions/MetaDataConfigurationActions.js
```diff
@@ -0,0 +1,47 @@
...
+var MetaDataConfigurationActionTypes = require("../constants/MetaDataConfigurationActionTypes");
+var AppEventEmitter = require("../EventEmitter");
+var NodeEventEmitter = require("events").EventEmitter;
+var APIResponseMixin = require("../mixins/APIResponseMixin");
+var update = require("react-addons-update");
+
+class MetaDataConfigurationActions extends NodeEventEmitter {
+  changeField(fieldName, fieldValues, pushToUrl) {
+    // Clone values in order to revert the store if the change fails
+    const previousValues = update(MetaDataConfigurationStore.fields[fieldName], {});
```
jgondron: This is defined in our components.js

jonhartzler:
:+1:

PR #320, opened 2016-02-08
DEC-662: Update beehive work with site_path

jgondron:
Why: Before I can work on beehive, we need to change site_objects to site_path in honeycomb.
How: Migrated site_objects to site_path and updated all references to site_path.

Renamed branch from https://github.com/ndlib/honeycomb/pull/319


jonhartzler:
:+1:

rfox2:
:+1:

PR #321, opened 2016-02-09
Dec 766 allow metaconfig to be updated

jonhartzler:
Here I updated the configuration to have an update service.  This is the first part of DEC-766.  The second part is coming soon!!

### DIFF
--- a/app/controllers/v1/configurations_controller.rb
+++ b/app/controllers/v1/configurations_controller.rb
@@ -1,0 +1,13 @@
+module V1
+  class ConfigurationsController < APIController
+    def show
+      collection = CollectionQuery.new.any_find(params[:collection_id])
+      @configuration = CollectionConfigurationQuery.new(collection).find
+      @configuration = V1::MetadataConfigurationJSON.new(@configuration)
+
+      respond_to do |format|
+        format.json { render json: @configuration.to_json }
+      end
+    end
+  end
+end
--- a/app/controllers/v1/metadata_fields_controller.rb
+++ b/app/controllers/v1/metadata_fields_controller.rb
@@ -1,0 +1,25 @@
+module V1
+  class MetadataFieldsController < APIController
+    def index
+      collection = CollectionQuery.new.any_find(params[:collection_id])
+      @configuration = CollectionConfigurationQuery.new(collection).find
+      @configuration = V1::MetadataConfigurationJSON.new(@configuration)
+
+      respond_to do |format|
+        format.json { render json: @configuration.to_json }
+      end
+    end
+
+    def update
+      @collection = CollectionQuery.new.any_find(params[:collection_id])
+
+      return if rendered_forbidden?(@collection)
+
+      @return_value = Metadata::UpdateConfigurationField.call(@collection, params[:field])
+
+      respond_to do |format|
+        format.json { render json: { status: @return_value }.to_json }
+      end
+    end
+  end
+end
--- a/app/controllers/v1/metadata_fields_controller.rb
+++ b/app/controllers/v1/metadata_fields_controller.rb
@@ -1,21 +1,11 @@
 module V1
   class MetadataFieldsController < APIController
-    def index
-      collection = CollectionQuery.new.any_find(params[:collection_id])
-      @configuration = CollectionConfigurationQuery.new(collection).find
-      @configuration = V1::MetadataConfigurationJSON.new(@configuration)
-
-      respond_to do |format|
-        format.json { render json: @configuration.to_json }
-      end
-    end
-
     def update
       @collection = CollectionQuery.new.any_find(params[:collection_id])
 
       return if rendered_forbidden?(@collection)
 
-      @return_value = Metadata::UpdateConfigurationField.call(@collection, params[:field])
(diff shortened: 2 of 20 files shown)


app/controllers/v1/metadata_fields_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+  class MetadataFieldsController < APIController
+    def update
+      @collection = CollectionQuery.new.any_find(params[:collection_id])
+
+      return if rendered_forbidden?(@collection)
+
+      @return_value = Metadata::UpdateConfigurationField.call(@collection, params[:id], params[:fields])
+
+      respond_to do |format|
+        format.json { render json: { status: @return_value }.to_json }
```
jgondron: This needs to render unprocessable_entity with the errors similar to the way v1/items_controller#update works, then the dialog should notify the user of the problematic fields.

app/controllers/v1/configurations_controller.rb
```diff
@@ -0,0 +1,13 @@
+module V1
+  class ConfigurationsController < APIController
+    def show
+      collection = CollectionQuery.new.any_find(params[:collection_id])
+      @configuration = CollectionConfigurationQuery.new(collection).find
+      @configuration = V1::MetadataConfigurationJSON.new(@configuration)
```
rfox2: Didn't you DRY this up in a previous merge?

app/models/metadata/configuration.rb
```diff
@@ -15,6 +11,18 @@ def initialize(data)
...
 
+    def save_field(name, values)
+      f = field(name)
+      return false if !f
+
+      if !f.update(values)
+        return false
+      end
+
+      data.metadata = fields.map(&:to_hash)
```
rfox2: Does this just take each field and call to_hash on it?

app/models/metadata/configuration/field.rb
```diff
@@ -1,9 +1,13 @@
...
   class Configuration
     class Field
+      include ActiveModel::Validations
       TYPES = [:string, :html, :date]
 
-      attr_reader :name, :type, :label, :multiple, :required, :default_form_field, :optional_form_field, :order, :placeholder, :help, :boost
+      attr_accessor :name, :type, :label, :multiple, :required, :default_form_field, :optional_form_field, :order, :placeholder, :help, :boost
+
+      validates :name, :type, :label, :order, presence: true
+      validates :type, inclusion: TYPES
```
rfox2: I think I know what this means, but does this limit the type field to the symbols listed in the constant array?

app/models/metadata/configuration/field.rb
```diff
@@ -36,10 +40,59 @@ def initialize(
...
+      private
+
+      def convert_keys_to_json!(json)
         json["defaultFormField"] = json.delete("default_form_field")
         json["optionalFormField"] = json.delete("optional_form_field")
+      end
 
-        json
+      def convert_json_to_ruby_keys!(hash)
+        hash.to_hash.symbolize_keys!
```
rfox2: I think I asked this before, but you need to do this in order to disambiguate the type of key that we use correct? (symbol vs. string)

app/models/metadata/configuration/field.rb
```diff
@@ -36,10 +40,59 @@ def initialize(
...
+        hash[:type] = hash[:type].to_sym if hash[:type]
```
rfox2: Follow up question to the one above. If you call symbolize_keys! on the resultant hash, why do you need to call to_sym on this...is it to symbolize the value? I think I'm answering my own question.

app/controllers/v1/configurations_controller.rb
jonhartzler: This the dryed version notice that line 6 is the json decorator.  in the previous version there was another line between 5 and 6

app/models/metadata/configuration.rb
jonhartzler: yes it is returning the data to the jsonb store.

rfox2:
Just some curiosity questions but other than that I'm :+1: to go with this! (after Justin's concern is addressed)

app/models/metadata/configuration/field.rb
jonhartzler: yes

app/models/metadata/configuration/field.rb
jonhartzler: yes this allows my passed in hashes for updates to be either symobls or strings.  
.to_hash converts the ActionController::Params object which is what you get when you use a controller to pass new params in to an actual hash so it can go into the jsonb field.

app/models/metadata/configuration/field.rb
jonhartzler: because to pass validation it must be a symbol.  since that is all the validation is allowing.

rfox2: I didn't know that. Very good to know.

app/controllers/v1/metadata_fields_controller.rb
jonhartzler: This is in the next pull request.

jgondron:
:+1:

PR #323, opened 2016-02-17
Dec 766 allow metaconfig to be created

jonhartzler:
Allow the creation of metadata fields.  

1.  Updates to the form to create fields.
2.  Added a parameter normalizer to ensure data is saved correctly. 
3.  Added routes and services to create

### DIFF
--- a/app/assets/javascripts/actions/MetaDataConfigurationActions.js
+++ b/app/assets/javascripts/actions/MetaDataConfigurationActions.js
@@ -6,7 +6,7 @@ var APIResponseMixin = require("../mixins/APIResponseMixin");
 var update = require("react-addons-update");
 
 class MetaDataConfigurationActions extends NodeEventEmitter {
-  changeField(fieldName, fieldValues, pushToUrl) {
+  changeField(fieldName, fieldValues, pushToUrl, update) {
     // Clone values in order to revert the store if the change fails
     const previousValues = update(MetaDataConfigurationStore.fields[fieldName], {});
     // Optimistically change the store
--- a/app/assets/javascripts/actions/MetaDataConfigurationActions.js
+++ b/app/assets/javascripts/actions/MetaDataConfigurationActions.js
@@ -6,7 +6,7 @@ var APIResponseMixin = require("../mixins/APIResponseMixin");
 var update = require("react-addons-update");
 
 class MetaDataConfigurationActions extends NodeEventEmitter {
-  changeField(fieldName, fieldValues, pushToUrl, update) {
+  changeField(fieldName, fieldValues, pushToUrl, create) {
     // Clone values in order to revert the store if the change fails
     const previousValues = update(MetaDataConfigurationStore.fields[fieldName], {});
     // Optimistically change the store
@@ -16,11 +16,14 @@ class MetaDataConfigurationActions extends NodeEventEmitter {
       values: fieldValues,
     });
 
-    pushToUrl += "/" + fieldName;
+    if (!create) {
+      pushToUrl += "/" + fieldName;
+    }
+
     $.ajax({
       url: pushToUrl,
       dataType: "json",
-      method: "PUT",
+      method: (create) ? "POST" : "PUT",
       data: {
         fields: fieldValues,
       },
@@ -30,12 +33,14 @@ class MetaDataConfigurationActions extends NodeEventEmitter {
         AppEventEmitter.emit("MessageCenterDisplay", "info", "Collection updated");
       }).bind(this),
       error: (function(xhr) {
-        // Request to change failed, revert the store to previous values
-        AppDispatcher.dispatch({
-          actionType: MetaDataConfigurationActionTypes.MDC_CHANGE_FIELD,
-          name: fieldName,
-          values: previousValues,
-        });
+        if (!create) {
+          // Request to change failed, revert the store to previous values on if it is updated
+          AppDispatcher.dispatch({
+            actionType: MetaDataConfigurationActionTypes.MDC_CHANGE_FIELD,
+            name: fieldName,
+            values: previousValues,
+          });
+        }
         // Communicate the error to the user
         this.emit("ChangeFieldFinished", false, xhr);
         AppEventEmitter.emit("MessageCenterDisplay", "error", APIResponseMixin.apiErrorToString(xhr));
--- a/app/assets/javascripts/components/forms/DateField.jsx
+++ b/app/assets/javascripts/components/forms/DateField.jsx
@@ -37,17 +37,18 @@ var DateField = React.createClass({
     if (!this.props.value) {
       return;
     }
-    if (this.props.value.length) {
-      this.props.value = this.props.value[0];
+    var value = this.props.value;
+    if (value.length) {
+      value = value[0];
     }
 
     this.setState({
-      year: this.props.value.year,
-      month: this.props.value.month,
-      day: this.props.value.day,
-      bc: this.props.value.bc,
-      displayText: this.props.value.display_text,
-      chooseDisplayText: (this.props.value.display_text ? true : false)
+      year: value.year,
+      month: value.month,
+      day: value.day,
+      bc: value.bc,
+      displayText: value.display_text,
+      chooseDisplayText: (value.display_text ? true : false)
     });
   },
 
--- a/app/assets/javascripts/components/forms/MetaDataFieldDialog.jsx
+++ b/app/assets/javascripts/components/forms/MetaDataFieldDialog.jsx
@@ -15,7 +15,14 @@ var MetaDataFieldDialog = React.createClass({
   propTypes: {
     open: React.PropTypes.bool.isRequired,
     updateUrl: React.PropTypes.string.isRequired,
+    createForm: React.PropTypes.bool,
     fieldName: React.PropTypes.string,
+  },
+
+  getDefaultProps: function() {
+    return {
+      createForm: false
+    }
   },
 
   // Creating custom react link so that we can store the field values
@@ -32,6 +39,10 @@ var MetaDataFieldDialog = React.createClass({
   updateFieldValue: function(key, value) {
     var kvp = {};
     kvp[key] = value;
+    if (key == "required") {
+      kvp["defaultFormField"] = true
+    }
+
     var fieldValues = update(this.state.fieldValues, {$merge: kvp});
     this.setState({ fieldValues: fieldValues });
   },
@@ -41,6 +52,7 @@ var MetaDataFieldDialog = React.createClass({
       open: this.props.open,
       fieldName: this.props.fieldName,
       saving: false,
+      createForm: false,
     };
   },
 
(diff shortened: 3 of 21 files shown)


app/assets/javascripts/actions/MetaDataConfigurationActions.js
```diff
@@ -16,11 +16,14 @@ class MetaDataConfigurationActions extends NodeEventEmitter {
...
-    pushToUrl += "/" + fieldName;
+    if (!create) {
+      pushToUrl += "/" + fieldName;
+    }
+
     $.ajax({
       url: pushToUrl,
       dataType: "json",
-      method: "PUT",
+      method: (create) ? "POST" : "PUT",
```
rfox2: Just an observation. I like this use of the ternary here.

app/assets/javascripts/actions/MetaDataConfigurationActions.js
```diff
@@ -30,12 +33,14 @@ class MetaDataConfigurationActions extends NodeEventEmitter {
...
-          actionType: MetaDataConfigurationActionTypes.MDC_CHANGE_FIELD,
-          name: fieldName,
-          values: previousValues,
-        });
+        if (!create) {
+          // Request to change failed, revert the store to previous values on if it is updated
+          AppDispatcher.dispatch({
+            actionType: MetaDataConfigurationActionTypes.MDC_CHANGE_FIELD,
+            name: fieldName,
+            values: previousValues,
```
rfox2: Will this always be an array?

app/assets/javascripts/components/forms/DateField.jsx
```diff
@@ -37,17 +37,18 @@ var DateField = React.createClass({
     if (!this.props.value) {
       return;
     }
-    if (this.props.value.length) {
-      this.props.value = this.props.value[0];
+    var value = this.props.value;
+    if (value.length) {
```
rfox2: At this point, why would this be an array?

jonhartzler: All metadata are stored as arrays to normalize the storage of multiple vs single fields.  
"name" => ["name"]

app/assets/javascripts/components/forms/MetaDataFieldDialog.jsx
```diff
@@ -125,7 +174,7 @@ var MetaDataFieldDialog = React.createClass({
       <div>
         <Dialog
           ref="EditMetaDialog"
-          title="Edit Metadata Field"
+          title={this.title()}
```
rfox2: Are we absolutely sure there will always be a title value?

app/controllers/v1/items_controller.rb
```diff
@@ -59,27 +59,7 @@ def showcases
...
-        [:contributor, contributor: []], # both :contributor, and contributor: [] are required bc it can pass null or an array.
-        [:original_language, original_language: []],
-        [:subject, subject: []],
-        [:call_number, call_number: []],
-        [:provenance, provenance: []],
-        [:date_created, [:value, :year, :month, :day, :bc, :display_text]],
-        [:date_modified, [:value, :year, :month, :day, :bc, :display_text]],
-        [:date_published, [:value, :year, :month, :day, :bc, :display_text]],
-      )
+      params[:item].to_hash
```
rfox2: Nice to use a straight hash at this point, isn't it? ;-)

app/assets/javascripts/actions/MetaDataConfigurationActions.js
jonhartzler: in general yes.  all values are stored as arrays.

app/controllers/v1/metadata_fields_controller.rb
```diff
@@ -5,7 +5,25 @@ def update
 
       return if rendered_forbidden?(@collection)
 
-      @return_value = Metadata::UpdateConfigurationField.call(@collection, params[:id], params[:fields])
+      @return_value = :success
+      if !Metadata::UpdateConfigurationField.call(@collection, params[:id], params[:fields])
+        @return_value = :unprocessable_entity
+      end
+
+      respond_to do |format|
```
rfox2: Curiosity question: if the API is always expected to serialize out with JSON, why do we need a respond_to here?

app/controllers/v1/items_controller.rb
jonhartzler: well we lose the parameter white listing that was going on at this point.

app/assets/javascripts/components/forms/MetaDataFieldDialog.jsx
jonhartzler: that is the idea.

app/controllers/v1/metadata_fields_controller.rb
jonhartzler: This is not serializing just saying yes or no to if it worked.

app/services/metadata/configuration_input_cleaner.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    def initialize(data)
+      @data = data
+    end
+
+    def clean!
+      ensure_symbols
+      ensure_json_fields_are_converted
+      ensure_name_from_label
```
rfox2: I'm not sure why this needs to be sanitized this way. Is it this way to avoid injection attacks? The label will be used as the key right?

jonhartzler: when you create a new metadata field we do not allow them to set a name value.  We create it for them.

app/services/metadata_input_cleaner.rb
```diff
@@ -12,7 +12,7 @@ def initialize(item)
   def clean!
     item.metadata.each do |key, value|
       if !value.is_a?(Array)
-        item.metadata[key] = [value]
+        item.metadata[key] = value["0"] ? value["0"] : [value]
```
rfox2: Why string values for array indexes?

jonhartzler: dates come through this way from react and the ajax post.  I tried many ways to get it to act like a real array but it always wants to be a hash.

spec/services/metatdata/configuration_input_cleaner_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  it "downcases labels into the name " do
+    data[:label] = "Name"
+    expect(subject).to eq(label: "Name", name: "name")
+  end
+
+  it "underscores spaces in the label to names" do
+    data[:label] = "na me"
+    expect(subject).to eq(label: "na me", name: "na_me")
+  end
+
```
rfox2: Should there be specs for ensure_name_from_label()?

app/assets/javascripts/components/forms/MetaDataFieldDialog.jsx
rfox2: Do we validate that along the way?

spec/services/metatdata/configuration_input_cleaner_spec.rb
jonhartzler: I did not write one because this is basically testing it already and if the name ever stops getting generated from the label both of these tests will fail.

app/controllers/v1/metadata_fields_controller.rb
rfox2: No, I mean the respond_to (line 13). Is that necessary?

app/services/metadata/configuration_input_cleaner.rb
rfox2: If we do that automatically, why do we need to sanitize the input? I've no problem with that, just curious if it's necessary.

app/services/metadata_input_cleaner.rb
rfox2: That's a pain. :-)

app/controllers/v1/items_controller.rb
rfox2: Gotcha

app/assets/javascripts/actions/MetaDataConfigurationActions.js
rfox2: That answers my other questions along this line.

app/services/metadata/configuration_input_cleaner.rb
jonhartzler: this is us doing it automatically.

app/controllers/v1/metadata_fields_controller.rb
jonhartzler: perhaps not it was copied from another controller.  I like it because it does indicate clearly what responses are available.

app/services/metadata/configuration_input_cleaner.rb
rfox2: Oh, I see. This isn't just another layer of input validation then. Gotcha.

spec/services/metatdata/configuration_input_cleaner_spec.rb
rfox2: Hmmm. I know what you mean but I'd like an explicit spec for that. You can slap me when you come in tomorrow.

app/assets/javascripts/components/forms/MetaDataFieldDialog.jsx
jonhartzler: this is calling 

  title: function() {
    return this.state.createForm ? "New Metadata Field" : "Edit Metadata Field";
  },

it is not referencing an item title.

rfox2:
When the additional test is written, :+1:

app/assets/javascripts/stores/MetaDataConfigurationStore.js
```diff
@@ -24,6 +24,10 @@ class MetaDataConfigurationStore extends EventEmitter {
   }
 
   changeField(name, values) {
+    // ensure the values
+    if (!values.order) {
```
jgondron: I think populating a reasonable default for order would be better handled on the back end. It could do a query for max + 1. Right now our order doesn't match 1-1 with size (user_defined_id order is 100).

jonhartzler: It does both at the moment and I figured that part of writing an order algorith would be to ensure the order was correct.  I am also migrating the user defined id in next pull request.

app/controllers/v1/items_controller.rb
jgondron: Could we still whitelist params based on if they exist in the collection's meta configs?

app/controllers/v1/metadata_fields_controller.rb
```diff
@@ -5,7 +5,25 @@ def update
...
+        format.json { render json: { status: @return_value }.to_json }
```
jgondron: This should render validation errors similar to v1::ItemsController when status is unprocessable_entity. Otherwise we won't be able to tell the user why the save failed.

app/controllers/v1/metadata_fields_controller.rb
```diff
@@ -5,7 +5,25 @@ def update
...
+
+    def create
+      @collection = CollectionQuery.new.any_find(params[:collection_id])
+
+      return if rendered_forbidden?(@collection)
+
+      @return_value = :success
+      if !Metadata::CreateConfigurationField.call(@collection, params[:fields])
+        @return_value = :unprocessable_entity
+      end
```
jgondron: This should render validation errors similar to v1::ItemsController when status is unprocessable_entity. Otherwise we won't be able to tell the user why the save failed.

jgondron:
:+1:

PR #324, opened 2016-02-18
DEC-796: User would like to remove/restore metadata fields

jgondron:
Why: User needs a way to remove and restore removed fields from their configuration, except for necessary fields like name.
How: 
  - Added an "active" property to all fields. 
  - Changed the configuration form to a list with a button that allows them to remove/restore each field. 
  - Added an immutable property list to each metadata field that allows us to restrict what properties a user can change. Ex: for the name field, active is now an immutable property.
  - Changed item api to filter out metadata fields that are inactive.
  - Changed search to filter out inactive fields from facets and sorts.

### DIFF
--- a/app/assets/javascripts/actions/MetaDataConfigurationActions.js
+++ b/app/assets/javascripts/actions/MetaDataConfigurationActions.js
@@ -12,7 +12,7 @@ class MetaDataConfigurationActions extends NodeEventEmitter {
     var fieldValues = update(MetaDataConfigurationStore.fields[fieldName], {});
 
     // Optimistically change the store
-    fieldValues["active"] = activeValue;
+    fieldValues.active = activeValue;
     AppDispatcher.dispatch({
       actionType: MetaDataConfigurationActionTypes.MDC_CHANGE_FIELD,
       name: fieldName,
@@ -34,7 +34,7 @@ class MetaDataConfigurationActions extends NodeEventEmitter {
       }).bind(this),
       error: (function(xhr) {
         // Request to change failed, revert the store to previous values
-        fieldValues["active"] = previousValue;
+        fieldValues.active = previousValue;
         AppDispatcher.dispatch({
           actionType: MetaDataConfigurationActionTypes.MDC_CHANGE_FIELD,
           name: fieldName,
--- a/app/assets/javascripts/actions/MetaDataConfigurationActions.js
+++ b/app/assets/javascripts/actions/MetaDataConfigurationActions.js
@@ -6,6 +6,47 @@ var APIResponseMixin = require("../mixins/APIResponseMixin");
 var update = require("react-addons-update");
 
 class MetaDataConfigurationActions extends NodeEventEmitter {
+  changeActive(fieldName, activeValue, pushToUrl){
+    // Clone values in order to revert the store if the change fails
+    const previousValue = MetaDataConfigurationStore.fields[fieldName].active;
+    var fieldValues = update(MetaDataConfigurationStore.fields[fieldName], {});
+
+    // Optimistically change the store
+    fieldValues["active"] = activeValue;
+    AppDispatcher.dispatch({
+      actionType: MetaDataConfigurationActionTypes.MDC_CHANGE_FIELD,
+      name: fieldName,
+      values: fieldValues,
+    });
+
+    pushToUrl += "/" + fieldName;
+    $.ajax({
+      url: pushToUrl,
+      dataType: "json",
+      method: "PUT",
+      data: {
+        fields: fieldValues,
+      },
+      success: (function() {
+        // Store was already changed, nothing to do here
+        this.emit("ChangeActiveFinished", true);
+        AppEventEmitter.emit("MessageCenterDisplay", "info", "Collection updated");
+      }).bind(this),
+      error: (function(xhr) {
+        // Request to change failed, revert the store to previous values
+        fieldValues["active"] = previousValue;
+        AppDispatcher.dispatch({
+          actionType: MetaDataConfigurationActionTypes.MDC_CHANGE_FIELD,
+          name: fieldName,
+          values: fieldValues,
+        });
+        // Communicate the error to the user
+        this.emit("ChangeActiveFinished", false, xhr);
+        AppEventEmitter.emit("MessageCenterDisplay", "error", APIResponseMixin.apiErrorToString(xhr));
+      }).bind(this)
+    });
+  }
+
   changeField(fieldName, fieldValues, pushToUrl) {
     // Clone values in order to revert the store if the change fails
     const previousValues = update(MetaDataConfigurationStore.fields[fieldName], {});
--- a/app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
+++ b/app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
@@ -39,7 +39,7 @@ var ItemMetaDataForm = React.createClass({
 
   getInitialState: function() {
     return {
-      formFields: _.sortBy(MetaDataConfigurationStore.fields, 'order'),
+      formFields: _.sortBy(MetaDataConfigurationStore.activeFields, 'order'),
       formValues: this.props.data,
       formState: "new",
       dataState: "clean",
@@ -69,7 +69,7 @@ var ItemMetaDataForm = React.createClass({
 
   setFormFieldsFromConfiguration: function() {
     this.setState({
-      formFields: _.sortBy(MetaDataConfigurationStore.fields, 'order'),
+      formFields: _.sortBy(MetaDataConfigurationStore.activeFields, 'order'),
     });
   },
 
--- a/app/assets/javascripts/components/forms/MetadataConfigurationForm.jsx
+++ b/app/assets/javascripts/components/forms/MetadataConfigurationForm.jsx
@@ -1,5 +1,16 @@
-var React = require('react');
+var React = require("react");
 var mui = require("material-ui");
+
+var ReactLink = require('react/lib/ReactLink');
+var ReactStateSetters = require('react/lib/ReactStateSetters');
+var MetaDataConfigurationActions = require("../../actions/MetaDataConfigurationActions");
+var Paper = mui.Paper;
+var List = mui.List;
+var FontIcon = mui.FontIcon;
+var IconButton = mui.IconButton;
+var Toggle = mui.Toggle;
+var Colors = require("material-ui/lib/styles/colors");
+var MoreVertIcons = require("material-ui/lib/svg-icons/navigation/more-vert");
 
 var MetaDataConfigurationForm = React.createClass({
   propTypes: {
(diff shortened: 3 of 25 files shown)


app/assets/javascripts/actions/MetaDataConfigurationActions.js
```diff
@@ -6,6 +6,47 @@ var APIResponseMixin = require("../mixins/APIResponseMixin");
 var update = require("react-addons-update");
 
 class MetaDataConfigurationActions extends NodeEventEmitter {
+  changeActive(fieldName, activeValue, pushToUrl){
+    // Clone values in order to revert the store if the change fails
+    const previousValue = MetaDataConfigurationStore.fields[fieldName].active;
```
jgondron: False

app/assets/javascripts/stores/MetaDataConfigurationStore.js
```diff
@@ -10,6 +10,7 @@ class MetaDataConfigurationStore extends EventEmitter {
     this._data = {};
 
     Object.defineProperty(this, "fields", { get: function() { return this._data.fields; } });
+    Object.defineProperty(this, "activeFields", { get: function() { return _.where(this._data.fields, {active: true}); } });
```
jgondron: False

app/assets/javascripts/actions/MetaDataConfigurationActions.js
```diff
@@ -6,6 +6,47 @@ var APIResponseMixin = require("../mixins/APIResponseMixin");
...
+    var fieldValues = update(MetaDataConfigurationStore.fields[fieldName], {});
```
jgondron: False

jonhartzler:
This looks good but I am not seeing specs on the changes in sort.rb, search/facet.rb or configuration/facet.rb

jonhartzler:
:+1:

app/assets/javascripts/actions/MetaDataConfigurationActions.js
rfox2: Yeah, why does hound do this? I've had this happen to me a few times and gone "huh? - it's defined right there!"

app/assets/javascripts/actions/MetaDataConfigurationActions.js
```diff
@@ -6,6 +6,47 @@ var APIResponseMixin = require("../mixins/APIResponseMixin");
...
+      actionType: MetaDataConfigurationActionTypes.MDC_CHANGE_FIELD,
+      name: fieldName,
+      values: fieldValues,
+    });
+
+    pushToUrl += "/" + fieldName;
+    $.ajax({
+      url: pushToUrl,
+      dataType: "json",
+      method: "PUT",
```
rfox2: Is this component only responsible for updating existing metadata field values?

app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
```diff
@@ -39,7 +39,7 @@ var ItemMetaDataForm = React.createClass({
 
   getInitialState: function() {
     return {
-      formFields: _.sortBy(MetaDataConfigurationStore.fields, 'order'),
+      formFields: _.sortBy(MetaDataConfigurationStore.activeFields, 'order'),
```
rfox2: Are the active fields the fields that they have "switched on"?

app/assets/javascripts/actions/MetaDataConfigurationActions.js
jgondron: This method only changes the active property for existing fields, but the component as a whole updates and creates. The changeField method allows creating new fields

app/assets/javascripts/components/forms/MetadataConfigurationForm.jsx
```diff
@@ -1,6 +1,18 @@
-var React = require('react');
+var React = require("react");
 var mui = require("material-ui");
 
+var ReactLink = require('react/lib/ReactLink');
+var ReactStateSetters = require('react/lib/ReactStateSetters');
+var MetaDataConfigurationActions = require("../../actions/MetaDataConfigurationActions");
+var Paper = mui.Paper;
```
rfox2: So, this is just a gripe but I wish we didn't have to explicitly include the MUI components that the component uses. I can see arguments for either way: explicit inclusion or implicit but if we're using MUI it sort of seems like it's OK for that to be implicit.

app/assets/javascripts/components/forms/MetadataConfigurationForm.jsx
```diff
@@ -8,23 +20,30 @@ var MetaDataConfigurationForm = React.createClass({
...
     return {
-      fields: this.sortedFields(),
+      fields: this.filteredFields(false),
       selectedField: undefined,
+      showInactive: false,
     };
   },
 
-  componentWillMount: function() {
+  componentDidMount: function() {
```
rfox2: This is using Flux right? So what you're saying here is that after the component mounts, check the MetaDataConfigurationStoreChanged attr on the flux store? Just trying to understand the flow.

app/assets/javascripts/components/forms/MetadataConfigurationForm.jsx
```diff
@@ -41,25 +60,108 @@ var MetaDataConfigurationForm = React.createClass({
     }
   },
 
-  getFieldRows: function() {
+  getLeftIcon: function(type) {
+    switch(type){
```
rfox2: Where's the "are you some sort of crazy academic for including this??" icon? I think we need that.

app/assets/javascripts/components/forms/MetadataConfigurationForm.jsx
```diff
@@ -41,25 +60,108 @@ var MetaDataConfigurationForm = React.createClass({
...
+  backgroundStyle: function() {
+    return {
+      maxWidth: "400px",
+      marginTop: "0px",
+      marginLeft: "48px",
+      padding: "20px"
+    };
+  },
+
+  listStyle: function() {
```
rfox2: This seems crazy. You're always going to return nil right?

app/assets/javascripts/components/forms/MetadataConfigurationForm.jsx
```diff
@@ -69,24 +171,14 @@ var MetaDataConfigurationForm = React.createClass({
   render: function(){
     const { selectedField } = this.state;
     return (
-      <div>
+      <Paper style={ this.backgroundStyle() } zDepth={0}>
```
rfox2: Why did you need the zDepth attr?

app/assets/javascripts/stores/MetaDataConfigurationStore.js
rfox2: If you don't define "_" explicitly, I reject all of the new code you've written here. All of it.

app/models/metadata/configuration/field.rb
```diff
@@ -68,8 +75,14 @@ def update(new_attributes)
         end
         convert_json_to_ruby_keys!(new_attributes)
 
+        if new_attributes.has_key?(:immutable)
```
rfox2: Why would an attr be "immutable"?

app/models/waggle/search/result.rb
```diff
@@ -24,14 +24,16 @@ def total
       end
 
       def facets
-        adapter_result.facets
+        adapter_result.facets.select(&:active)
```
rfox2: Could you explain this syntax to me? I'm not familiar with the ampersand syntax.

rfox2:
Questions

app/assets/javascripts/actions/MetaDataConfigurationActions.js
rfox2: Did you see the way John wrote a ternary to determine which HTTP verb should be used depending on the action being taken? For REST, POST is used for create operations and PUT for update operations.

app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
jgondron: Technically at the moment, it's fields they haven't switched off. Since we're doing a soft delete of meta fields, we have to filter those out.

app/assets/javascripts/components/forms/MetadataConfigurationForm.jsx
jgondron: In general, we've moved to explicit inclusion for libraries since these components are meant to require everything they need to operate (similar to the css discussion).

That being said, as far as these specific declarations (such as Paper = mui.Paper), it isn't required. Within the renders, I could just as easily have done <mui.Paper /> instead of making this local variable that references mui.Paper. It's a matter of taste I suppose, the render looks cleaner but your declarations at the top are a little more verbose. There are some that I've found do not come with the default material-ui require. All of those have explicit require's to the appropriate library within material-ui.

app/assets/javascripts/components/forms/MetadataConfigurationForm.jsx
jgondron: It is using Flux. "MetaDataConfigurationStoreChanged" is an event message that the store will emit anytime its data changes. This component is subscribing to that event so that it can update it's state anytime the store changes. In this case, if the store ever changes, this component needs to update its state based on the current list of active fields defined.

The getAll() method is what initializes the store. This technically should not be handled inside this component, but rather through an action. I just haven't gotten around to refactoring that part.

jgondron: Also, something to keep in mind is that flux is really just a design pattern. The idea is you don't directly manipulate a store (which is just data shared between multiple components). Instead you go through an actions object that makes changes to the backend (if needed) and the local store. Your components subscribe to events emitted by the store so that they can update their state when a change that they care about is made to the store. The actual library used here is just a standard node event emitter.

app/assets/javascripts/components/forms/MetadataConfigurationForm.jsx
jgondron: Yeah it was a testing thing. I have some other changes that need to be made to layout to account for the "New field" button, so I'll pull this out as part of those changes.

app/assets/javascripts/components/forms/MetadataConfigurationForm.jsx
jgondron: Paper by default has a depth of 1 to always give it a small shadow. It doesn't look right on this subform system we have in settings.

app/models/metadata/configuration/field.rb
jgondron: Ex: The name field should never be allowed to be removed since our application uses it heavily. Some properties of a field like this need to be protected from user changes.

app/models/waggle/search/result.rb
jgondron: Nor was I. It's filtering to array elements whose active property is true. It's a shorthand version that was suggested by my wise and learned lud (aka the hound).

http://ruby-doc.org/core-1.8.7/Symbol.html#method-i-to_proc

app/assets/javascripts/actions/MetaDataConfigurationActions.js
jgondron: Right, which is why this method only does a PUT since it only changes existing fields.

rfox2: Oh sorry. I missed the other ajax call with the ternary. My bad.

app/assets/javascripts/components/forms/ItemMetaDataForm.jsx
rfox2: OK

app/assets/javascripts/components/forms/MetadataConfigurationForm.jsx
rfox2: Gotcha. That's very helpful.

app/models/metadata/configuration/field.rb
rfox2: Ah, I see

app/models/waggle/search/result.rb
rfox2: Oh, wild. OK.

rfox2:
:+1:

PR #325, opened 2016-02-25
DEC-802: Create field should respond with the new values

jgondron:
Why: The backend should populate default values for properties such as order, and the front end needs to receive what values were used to create the new field in order to update its store.
How: 
-Changed field saves to return the field object on successful save. This propagates up to the controller and the controller will render the new field's properties to the json response. 
-Changed new fields to be active by default, and use max order + 1 as a default for order. 
-Changed the changeField action to no longer change the store optimistically. This should not happen for new fields since we don't have a name or an order until we get a response from the server. If this needs to happen for updates, then we should break this up into two actions.
-Made a few minor cosmetic tweaks to the meta config form.


jonhartzler:
:+1:

rfox2:
:+1:

PR #326, opened 2016-02-25
Dec 755 remove item interface to metadata

jonhartzler:
Changed so that there is no public interface to the metadata.
I added the validation based on the configuration back in.


app/services/metadata_input_cleaner.rb
```diff
@@ -11,8 +11,20 @@ def initialize(item)
...
+    end
+
+    item.metadata.stringify_keys!
+  end
+
+  private
+
+  def ensure_value_is_array(key, value)
+    if !value.is_a?(Array)
+      if value.nil?
```
jonhartzler: This is how I want this to look because of the number of characters.

app/decorators/v1/metadata_json.rb
```diff
@@ -8,7 +8,8 @@ def metadata
       {}.tap do |hash|
         object.item_metadata.fields.each do |key, value|
           field_config = configuration.field(key)
-          if field_config.present? && field_config.active
+
+          if field_config.present? && value && field_config.active
```
jgondron: What is the desired behavior here? I didn't see a test that matches this change. If the value is a boolean and false, won't this get excluded from the json? May not be a problem now, but it might be better to explicitly test for whatever condition instead of using truthiness to avoid future problems.

app/models/item.rb
```diff
@@ -45,13 +29,33 @@ class Item < ActiveRecord::Base
   enum image_status: { no_image: 0, image_processing: 1, image_ready: 2, image_unavailable: 3 }
 
   def name
-    item_metadata.field(:name).first.value
+    item_metadata.name
```
jgondron: I thought we were removing all of these from the Item interface? (Same comment for description)

app/models/metadata/retrieval.rb
```diff
@@ -19,11 +35,24 @@ def field(field)
     end
 
     def field?(field)
-      get(field).present?
+      configuration.field_names.include?(field.to_s)
+    end
+
+    def set_metadata(metadata)
```
jgondron: Should this class be setting metadata? Makes it more than just retrieval.

rfox2:
:+1: after Justin's comments are addressed

jgondron:
:+1:

PR #328, opened 2016-03-03
DEC-801: Fix export import

jgondron:
Why: Lots of changes to the way metadata is handled to accommodate the dynamic metadata, which broke export/import.
How: User defined id is no longer part of metadata, but is still needed for export/import. Added special handling to merge this with meta fields. Changed the parsing/rewriting now that all metadata will be stored as arrays. Changed import to use the new metadata setter service to assign metadata to the items. Also fixed a problem associated with postgres' ordering of keys that caused imports to appear as though an item changed when it had not (See DEC-812 for more info).

### DIFF
--- a/app/services/find_or_create_item.rb
+++ b/app/services/find_or_create_item.rb
@@ -60,6 +60,10 @@ class FindOrCreateItem
   private
 
   def update_props
+    if props[:metadata].present?
+      metadata = props.delete(:metadata)
+      Metadata::Setter.call(item, metadata)
+    end
     item.assign_attributes(props)
     CreateUniqueId.call(item)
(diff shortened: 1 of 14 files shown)


app/services/find_or_create_item.rb
```diff
@@ -60,6 +60,10 @@ def find_or_create_by(criteria:)
   private
 
   def update_props
```
jonhartzler: I am concerned that we have 2 code paths for item creation and management.  After looking through this pr and the code I believe this could be turned into 1.  What are your thoughts?

jonhartzler:
:+1:

PR #330, opened 2016-03-04
add file for vatican only metadata

jonhartzler:
I added an example vaitcan metadata file.  This file can be loaded into your collections metadata. 
It is required if you want to use any of the new pulls for zucchetto.
I imagine this file will change over time as we work through the development of zucchetto.  
To load:

```
# get the config of the collection you want.
collection_configuration = Collection.first.collection_configuration
# load the yml 
config = YAML.load_file(Rails.root.join("config/metadata/", "vatican_item.yml"))
# set and save
collection_configuration.metadata = config[:fields]
collection_configuration.save! 

```

### DIFF
--- a/config/metadata/vatican_item.yml
+++ b/config/metadata/vatican_item.yml
@@ -1,0 +1,153 @@
+---
+:fields:
+  - :name: :name
+    :active: true
+    :type: :string
+    :label: Name
+    :required: true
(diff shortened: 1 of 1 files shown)


rfox2:
:+1:

danthewolfe:
:+1:

PR #337, opened 2016-03-15
DEC-829: Creating a new collection fails

jgondron:
Refactored the pieces that were shared between the new collection form and the general settings form into its own partial and changed those two forms to render that partial. This structure changed when we added subforms to settings to accommodate the metadata configuration.


rfox2:
:+1:

jonhartzler:
:+1:

PR #338, opened 2016-03-15
DEC-832: Reactivate a metadata field fails

jgondron:
Fixed a comparison problem with convert_strings_to_boolean. If the kvp was already converted (ex: it went through ConfigurationInputCleaner), then the == "true" would fail for true values, and convert it to false


danthewolfe:
:+1:

PR #339, opened 2016-03-17
DEC-835: Cannot create new metafield

jgondron:
Why: After making a change (such as deactivating) a configuration field, subsequent calls to create a new field would fail.
How: 
-The configuration properties "boost" and "order" are assumed to be integers, but the config update is receiving these as strings. Changed the input cleaner to convert stringified integers such as "15", but not things like "15abc". 
-Changed the validation to ensure boost and order are integers. This way if anything gets passed other than a basic integer that's been stringified, it will throw a validation error back to the app that posted the update.
-Separated create/update actions so that create can stop passing values for the properties that have none, this way the api can properly perform validation on all data passed in.

### DIFF
--- a/app/assets/javascripts/actions/MetaDataConfigurationActions.js
+++ b/app/assets/javascripts/actions/MetaDataConfigurationActions.js
@@ -79,7 +79,6 @@ class MetaDataConfigurationActions extends NodeEventEmitter {
         // Communicate the error to the user
         this.emit("ChangeFieldFinished", false, xhr);
         AppEventEmitter.emit("MessageCenterDisplay", "error", "Save failed. Please ensure to select a label and a type and try again.");
-        //AppEventEmitter.emit("MessageCenterDisplay", "error", APIResponseMixin.apiErrorToString(xhr));
       }).bind(this)
     });
   }
@@ -108,7 +107,6 @@ class MetaDataConfigurationActions extends NodeEventEmitter {
         // Communicate the error to the user
         this.emit("CreateFieldFinished", false, xhr);
         AppEventEmitter.emit("MessageCenterDisplay", "error", "Create failed. Please ensure to select a label and a type and try again.");
-        //AppEventEmitter.emit("MessageCenterDisplay", "error", APIResponseMixin.apiErrorToString(xhr));
       }).bind(this)
     });
   }
--- a/app/assets/javascripts/actions/MetaDataConfigurationActions.js
+++ b/app/assets/javascripts/actions/MetaDataConfigurationActions.js
@@ -47,18 +47,16 @@ class MetaDataConfigurationActions extends NodeEventEmitter {
     });
   }
 
-  changeField(fieldName, fieldValues, pushToUrl, create) {
+  changeField(fieldName, fieldValues, pushToUrl) {
     // Clone values in order to revert the store if the change fails
     const previousValues = update(MetaDataConfigurationStore.fields[fieldName], {});
 
-    if (!create) {
-      pushToUrl += "/" + fieldName;
-    }
+    pushToUrl += "/" + fieldName;
 
     $.ajax({
       url: pushToUrl,
       dataType: "json",
-      method: (create) ? "POST" : "PUT",
+      method: "PUT",
       data: {
         fields: fieldValues,
       },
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/actions/MetaDataConfigurationActions.js
```diff
@@ -72,17 +70,45 @@ class MetaDataConfigurationActions extends NodeEventEmitter {
...
+        AppDispatcher.dispatch({
+          actionType: MetaDataConfigurationActionTypes.MDC_CHANGE_FIELD,
+          name: fieldName,
+          values: previousValues,
+        });
         // Communicate the error to the user
         this.emit("ChangeFieldFinished", false, xhr);
-        AppEventEmitter.emit("MessageCenterDisplay", "error", APIResponseMixin.apiErrorToString(xhr));
+        AppEventEmitter.emit("MessageCenterDisplay", "error", "Save failed. Please ensure to select a label and a type and try again.");
+        //AppEventEmitter.emit("MessageCenterDisplay", "error", APIResponseMixin.apiErrorToString(xhr));
```
rfox2: Did you mean to leave this in as a comment?

app/assets/javascripts/actions/MetaDataConfigurationActions.js
```diff
@@ -72,17 +70,45 @@ class MetaDataConfigurationActions extends NodeEventEmitter {
...
+          values: result.field,
+        });
+        this.emit("CreateFieldFinished", true, result.field);
+        AppEventEmitter.emit("MessageCenterDisplay", "info", "Collection updated");
+      }).bind(this),
+      error: (function(xhr) {
+        // Communicate the error to the user
+        this.emit("CreateFieldFinished", false, xhr);
+        AppEventEmitter.emit("MessageCenterDisplay", "error", "Create failed. Please ensure to select a label and a type and try again.");
+        //AppEventEmitter.emit("MessageCenterDisplay", "error", APIResponseMixin.apiErrorToString(xhr));
```
rfox2: Same question here

app/services/metadata/configuration_input_cleaner.rb
```diff
@@ -43,6 +44,18 @@ def convert_value_to_bool(value)
       end
     end
 
+    def ensure_int_values
+      [:order, :boost].each do |key|
+        if data.has_key?(key)
+          begin
+            data[key] = Integer(data[key])
+          rescue
+            # Leave the value as is to allow the problem to pass through to field validation
```
rfox2: I'm sure you did test this in the specs but I'm not seeing it. Is this tested implicitly? Should it be explicit?

spec/models/metadata/configuration/field_spec.rb
```diff
@@ -141,7 +141,16 @@
...
-      expect(subject).to have(1).errors_on(:order)
+      expect(subject).to have_at_least(1).errors_on(:order)
+      expect(subject.errors.messages[:order]).to include("can't be blank")
+    end
+
+    [:order, :boost].each do |key|
+      it "validates #{key} is an integer" do
+        expect(subject.update(key => nil)).to be(false)
+        expect(subject).to have_at_least(1).errors_on(key)
+        expect(subject.errors.messages[key]).to include("is not a number")
```
rfox2: Is this where the pass through validation is tested?

app/assets/javascripts/actions/MetaDataConfigurationActions.js
```diff
@@ -47,18 +47,16 @@ class MetaDataConfigurationActions extends NodeEventEmitter {
...
-    if (!create) {
-      pushToUrl += "/" + fieldName;
-    }
+    pushToUrl += "/" + fieldName;
 
     $.ajax({
       url: pushToUrl,
       dataType: "json",
-      method: (create) ? "POST" : "PUT",
+      method: "PUT",
```
rfox2: I guess it makes sense that this should always use the PUT verb since this is a field update, correct?

rfox2:
I don't have any major concerns, just some questions. :+1:

app/services/metadata/configuration_input_cleaner.rb
jgondron: This is tested with the "does not convert" tests within the order, boost block

spec/models/metadata/configuration/field_spec.rb
jgondron: Yep. If the cleaner isn't able to convert a very basic stringified integer to an int, then validation will fail because boost and order need to be int. These specs make sure that validation is in place.

app/assets/javascripts/actions/MetaDataConfigurationActions.js
jgondron: Correct

jonhartzler:
:+1:

PR #340, opened 2016-03-19
DEC-837: Cannot save metadata field

jgondron:
Problem was the Date class was being overridden by Metadata::Fields::Date, which prevents ConvertToRubyDate from instantiating a valid Date object. Only occurred on preproduction due to eager loading.


jonhartzler:
:+1:

rfox2:
:+1:

PR #342, opened 2016-03-21
DEC-815: Cannot delete a collection

jgondron:
Why: Throws an exception due to foreign key constraint on the associated collection configuration.
How: Added a destroy service for collection config and changed DestroyCollection to use this new service to delete the associated config on cascade delete.


rfox2:
:+1:  When I first looked at the constructor for Destroy::Collection I thought it seemed complex but the syntax is actually pretty clean.

PR #344, opened 2016-03-23
DEC-849: Cannot edit sitepath

jgondron:
JS interpreter was having trouble with the const function reference connectDragSource.


rfox2:
:+1:

jonhartzler:
:+1:

PR #345, opened 2016-03-24
DEC-864, DEC-840: Cannot insert image into page

jgondron:
DEC-864:
-Temporarily disabled the ability to shortcut creating items via the insert image tool. This does not work correctly when background processing is enabled.
DEC-840:
-Also found a fix to the redactor reverting the changes after inserting an image. The insert.html seems to work better, and there were a few calls that insert.node was missing compared to insert.html.


rfox2:
:+1:

PR #347, opened 2016-04-01
DEC-889 Update vatican yaml config

rfox2:
What: Update the vatican metadata config so that it matches the latest data set


jonhartzler:
:+1:

PR #349, opened 2016-04-07
Update solr query parser

jonhartzler:
Changes the Query Parser to edismax which lets us use and AND and OR.   

Also it is in 2 places eventually I would remove the call in the Result Class eventually.  It allows me to switch the query type with out restarting solr which is currently a little bit of a pain.


jgondron:
:+1:

PR #352, opened 2016-04-15
DEC-913: Change default search to OR (3.0)

jgondron:
Why: For our purposes, search needs to default to OR when given terms separated by space.
How: Changed the mm param to solr to 1. This effectively works as an OR operator between optional terms. Things like +term or -term, or "term AND term" are all still obeyed.

For more info on the mm param, see: [The mm (Minimum Should Match) Parameter](https://cwiki.apache.org/confluence/display/solr/The+DisMax+Query+Parser#TheDisMaxQueryParser-Themm(MinimumShouldMatch)Parameter)


rfox2:
:+1:

jonhartzler:
👍

PR #353, opened 2016-04-19
DEC-924: Bugfix Safari const

jgondron:
Why: A show stopping bug is preventing JS functionality from executing on Safari. The browser is reporting a 'const' error, which is preventing React and jQuery from initializing.
How: Replaced all const with var.


rfox2:
👍

jonhartzler:
👍

PR #354, opened 2016-04-20
DEC-930: Ambrosiana import

jgondron:
Why: It will be easier to import Ambrosiana if we can import the images from the current website.
How: Modified the image upload task to allow for http based files. Now also prints which files were successfully imported. This way, if we have to bail in the middle for any reason, we can exclude these from the full list when running again.


jonhartzler:
👍

rfox2:
👍

PR #355, opened 2016-04-26
Dec 925 fix for promise bug

rfox2:
What: A bug was discovered related to the use of the Axios javascript library, only on IE and specifically IE 11. IE does not recognize the Promise object as having been defined.
How: Added a new module to NPM called 'es6-promise', which provides a polyfill back port that forces IE 10 and later to recognize the Promise definition.

**Note**: IE 10 and later also no longer recognize the comment conditionals (e.g. '[if IE]') so compensating javascript can no longer be included in the head of the document to correct IE compatibility issues

**ALSO**: Added an override to force Webrick to bind to IP address 0.0.0.0 to make it easier to debug from local VMs accessing Webrick running in the native OS

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -29,4 +29,5 @@
 //= require menu-bar
 //= require flash
 //= require jquery.cookie
+require('es6-promise').polyfill();
 window.$ = window.jQuery = require('jquery');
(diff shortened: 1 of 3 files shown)


danthewolfe:
👍

jonhartzler:
👍

PR #356, opened 2016-04-26
DEC-934 Fix for changed external collection attr

rfox2:
What: The attributes for collections and thus external collections changed and the description attr was deprecated. That needed to be corrected for the related forms and lists to work related to external collections.
How: I changed the attribute reference to 'about' instead of 'description' and updated the appropriate code and view pages to reflect that.


danthewolfe:
👍

jgondron:
👍

PR #357, opened 2016-04-27
Viewing changes on a papertrail enabled object

jgondron:
A problem came up this morning that required researching the changes on an object. Our implementation of PaperTrail does not give us the actual changes (see https://github.com/airblade/paper_trail#diffing-versions). This migration will give us access to the changeset for each version


rfox2:
I'd argue for just using the method provided by paper trail itself instead of writing our own method. The paper trail method only requires that we add an additional field to the versions table and paper trail does the rest. See https://github.com/airblade/paper_trail#diffing-versions

This is probably something that we would want to enable in other applications using paper trail. I don't know why we aren't using it in other apps. It does add a bit more storage to the database, but not much more than what we already have with vanilla paper trail.

In the end, if other think it's important to add this service class, I can live with that.

jonhartzler:
I think I am in favor of using the paper trail method too.  Less cognitive load but I am open to talk about it.

jgondron:
I changed it. Just keep in mind this won't help for existing records, so we may have to rewrite this code if we ever have to look at the history prior to this change getting published (hopefully should be a slim chance).

jonhartzler:
👍

rfox2:
👍  Thanks for being flexible Justin.

PR #358, opened 2016-04-28
DEC-947: Change rake task to replace existing images

jgondron:
Changed to replace the images of items that have the same name as that image (such as from a previous import)

### DIFF
--- a/lib/tasks/adhoc_upload_images.rake
+++ b/lib/tasks/adhoc_upload_images.rake
@@ -39,7 +39,7 @@ namespace :adhoc do
     file = open(file_path)
     file_name = File.basename(file.respond_to?(:base_uri) ? file.base_uri.path : file.path)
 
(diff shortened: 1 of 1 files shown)


lib/tasks/adhoc_upload_images.rake
```diff
@@ -36,14 +36,23 @@ namespace :adhoc do
...
+    item = find_or_create(collection_id: collection_id, file_name: file_name)
     save_params = { uploaded_image: file, metadata: { name: file_name } }
     result = SaveItem.call(item, save_params)
     Index::Item.index!(item)
     file.close
     result
   end
+
+  # Assumes that any previous items in the collection were named based on the file name.
+  def find_or_create(collection_id:, file_name:)
```
jonhartzler: my only thought is maybe we don't want this method name to collide with the rails magical method names?

jonhartzler:
👍

rfox2:
👍

PR #359, opened 2016-05-02
DEC-928 Fix for Safari onClick bug for metadata button

rfox2:
What: There is a bug related to Safari in material-ui that short circuits the on click events for certain components and forces the user to click the element twice for the event to fire.
How: Instead of using onClick(), I changed the component in question so that it uses onMouseDown() coupled with onTouchStart() for touch screen devices.


jonhartzler:
👍

danthewolfe:
👍

PR #363, opened 2016-05-02
DEC-648: Honeycomb collection list with large collections is slow

jgondron:
Problem was with how it was finding a thumbnail to use, not the item count. Changed it to use the collection image if it has one. If none is available, it will find the first item that has a honeypot image via a query instead of a block test.


rfox2:
As soon as the hound syntax issues are fixed, 👍

jonhartzler:
👍

PR #364, opened 2016-05-02
Dec 927 fix cache expiration metadata fields

rfox2:
What: Cache was persisting for items when metadata fields were changed at the collection level. The metadata configuration was added as a changeable object to the item level caching methods on the V1 API.
How: Added the configuration as a potential mutable object in generating the cache keys. Also updated tests to account for the presence of the collection_configuration() on collection objects.


jgondron:
👍

jonhartzler:
👍

PR #366, opened 2016-05-03
DEC-925: Promise polyfill deploy bug

jgondron:
Moving the polyfill to components.js. Is causing problems with the asset precompile on preprod/prod.


jonhartzler:
👍

PR #367, opened 2016-05-03
DEC-927: Metadata field cache bug

jgondron:
Page/Section/Showcase#show all also render item metadata, so they need to check the metadata configuration id/updated_at when generating etag.


rfox2:
👍  ship it

PR #368, opened 2016-05-13
DEC-1008: Bug fix export large collection

jgondron:
Why: Exports of large collections take a very long time or fail entirely. Ex: A sample collection with 17 columns of metadata and 10,785 rows will take approx 30-40 minutes and then fails to upload to google. This needs to succeed and the time must be reduced.
How: 
1. A large portion of the time was being spent on the update_cells call for a worksheet object in the google_drive gem. Forked this gem and created a fix on https://github.com/ndlib/google-drive-ruby/tree/DEC-1008-Speedup-update-cells. This reduced the time for update_cells from ~20 minutes to < 1 second. 
2. The post to google still took a large amount of time and would eventually fail. To fix this I changed the worksheet update/save calls to be batched by row count, which will be configurable in the settings.yml. The ideal row count will likely vary based on how many metadata fields a collection has, but for the example collection 2000 per batch works well as a good starting point.

Overall with these changes, the 183k cell example now successfully exports and takes approximately 5-7 minutes instead of 30+.


rfox2:
So, we're going to deploy a branch you created off of google_drive master? 👍

PR #369, opened 2016-06-03
DEC-624: Speedup item list

jgondron:
Why: Navigating through items for large collections is extremely slow, and the browser fails to load most images due to OOM.
How: Refactored item list to use the search API instead of loading all data on the client into datatables. This has caused an enormous speedup, ex: A collection of 10k items previously took 2+ minutes to load, and now loads in fractions of a second. A few additional fixes to the search API came out of this that we were not previously aware of: 
- Sorting now works correctly for alpha numerics.
- We are now able to provide a search direction (asc/desc) in a similar syntax to solr, ex: "sort=name desc"
- Can now sort by last updated time (updated_at) in addition to the collections configured metadata fields.

Please pull this branch down and thoroughly test searching, paging, and sorting while under different throttling presets (including no connection)


rfox2:
Before this gets merged, there's a usability issue I'd like to see get addressed. The right and left navigation buttons for navigation look like they will take the user to the next page but they navigate to the beginning or end of the entire list respectively. I'm not sure if there are style guidelines for this. What I've typically seen for that type of navigation is a double arrow, and arrow with a line, or just the text "first" or "last".

jgondron:
@rfox2 Fixed the first/last page icons. Those were based on the Beehive pagination, so we should probably change them there too.

rfox2:
👍

jonhartzler:
👍

PR #370, opened 2016-06-06
DEC-1048 Added custom slug field for collections

rfox2:
What: Added a field to the collection model that can store a custom URL slug that the user creates
How: Added field to model, added service class that can set the slug value or raise the appropriate exception

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -14,6 +14,7 @@ class Collection < ActiveRecord::Base
   validates_attachment_content_type :image, content_type: /\Aimage\/.*\Z/
   validates_attachment_content_type :uploaded_image, content_type: /\Aimage\/.*\Z/
   validates :name_line_1, :unique_id, presence: true
+  validates :url_slug, uniqueness: true, allow_nil: true
 
   has_paper_trail
 
--- a/app/services/set_collection_slug.rb
+++ b/app/services/set_collection_slug.rb
@@ -1,0 +1,21 @@
+class SetCollectionSlug
+  attr_reader :collection, :slug
+
+  def self.call(collection, slug)
+    new(collection, slug).set_slug!
+  end
+
+  def initialize(collection, slug)
+    @collection = collection
+    @slug = slug
+  end
+
+  def set_slug!
+    collection.url_slug = CreateURLSlug.call(@slug)
+    if collection.valid?
+      collection.save
+    else
+      raise "Cannot set slug: collection record not valid"
+    end
+  end
+end
--- a/app/services/set_collection_slug.rb
+++ b/app/services/set_collection_slug.rb
@@ -13,7 +13,7 @@ class SetCollectionSlug
   def set_slug!
     collection.url_slug = CreateURLSlug.call(@slug)
     if collection.valid?
-      collection.save!
+      collection.save
     else
       false
     end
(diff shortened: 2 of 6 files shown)


app/services/set_collection_slug.rb
```diff
@@ -0,0 +1,21 @@
...
+    new(collection, slug).set_slug!
+  end
+
+  def initialize(collection, slug)
+    @collection = collection
+    @slug = slug
+  end
+
+  def set_slug!
+    collection.url_slug = CreateURLSlug.call(@slug)
```
jgondron: Are we ok with having this rewrite what the user specified? Alternative would be to perform a validation step and report to the user the required format.

rfox2: I'm OK with an override. The same thing is happening with the title to slug conversion. If they type in a valid slug, it will take that. What we could do is put example text in the input, and perhaps instructions on the UI to let them know what a valid slug should look like.

jgondron: Right, but in the case of title->slug the user doesn't really care about the slug that gets created, only the title. Here they're directly setting the slug, so magical conversions that happen behind the scenes without telling them might make for bad UX.

rfox2: I'm OK with sanitizing it for them. The flash message could tell them what slug was created, plus it will show on the page when they're done.

jgondron:
👍

app/services/set_collection_slug.rb
```diff
@@ -0,0 +1,21 @@
...
+    if collection.valid?
+      collection.save
+    else
+      raise "Cannot set slug: collection record not valid"
```
jonhartzler: Services should return false on failure not raise an error.   Errors should be reserved for actual failures of code.   If you want to know what happened you can ask collection.errors for the failure.

db/migrate/20160606174800_add_custom_slug_to_collections.rb
```diff
@@ -0,0 +1,6 @@
+class AddCustomSlugToCollections < ActiveRecord::Migration
+  def change
+    add_column :collections, :url_slug, :string
+    add_index :collections, :url_slug, unique: true
```
jonhartzler: In addition to this there should be a unique constraint on the collection class so that the errors hash gets filled out with the error.

rfox2: OK, I'll add that.

app/services/set_collection_slug.rb
rfox2: Done

app/services/set_collection_slug.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def initialize(collection, slug)
+    @collection = collection
+    @slug = slug
+  end
+
+  def set_slug!
+    collection.url_slug = CreateURLSlug.call(@slug)
+    if collection.valid?
+      collection.save!
```
jonhartzler: This will raise an error.  I think we might want +      collection.save

app/models/collection.rb
```diff
@@ -14,6 +14,7 @@ class Collection < ActiveRecord::Base
   validates_attachment_content_type :image, content_type: /\Aimage\/.*\Z/
   validates_attachment_content_type :uploaded_image, content_type: /\Aimage\/.*\Z/
   validates :name_line_1, :unique_id, presence: true
+  validates :url_slug, uniqueness: true, allow_nil: true
```
jonhartzler: Is it possible to add a test for this in the spec?

rfox2: Done

jonhartzler:
👍

PR #371, opened 2016-06-08
DEC-1049 Form field for configuring the url slug

rfox2:
What: Added a form field for configuring the collection custom url slug. The form reports back if the system detects a duplicate.

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -43,6 +43,9 @@ class CollectionsController < ApplicationController
   def settings_update
     @collection = CollectionQuery.new.find(params[:id])
     check_user_edits!(@collection)
+    if params[:collection][:url_slug]
+      SetCollectionSlug.call(@collection, params[:collection][:url_slug])
+    end
 
     if SaveCollection.call(@collection, save_params)
       flash[:notice] = t(".success")
--- a/app/views/collections/_general_settings_subform.html.erb
+++ b/app/views/collections/_general_settings_subform.html.erb
@@ -2,4 +2,5 @@
 
 <%= f.input :name_line_1 %>
 <%= f.input :name_line_2 %>
+<%= f.input :url_slug, label: "Custom URL Path", hint: "ex: my-collection will become collections.library.nd.edu/my-collection" %>
 <%= render partial: 'search_and_browse_form', locals: { f: f } %>


app/controllers/collections_controller.rb
```diff
@@ -43,6 +43,9 @@ def settings
   def settings_update
     @collection = CollectionQuery.new.find(params[:id])
     check_user_edits!(@collection)
+    if params[:collection][:url_slug]
+      SetCollectionSlug.call(@collection, params[:collection][:url_slug])
```
jonhartzler: This probably should be tied the if statement for success below.

rfox2: Now that I'm thinking through the logic here I'm not happy about this either. The SetCollectionSlug service calls collection.save...now I'm wondering if all it should do is sanitize the slug string and set it on the collection object. Does that sound like a better solution? Then it would make sense to have the call come before the save below. Maybe the SetCollectionSlug service class should be refactored as a part of this PR, and I can commit that change separately.

jgondron: I agree, SetCollectionSlug should not save.

jgondron:
Also needs specs

danthewolfe:
👍

app/services/save_collection.rb
```diff
@@ -14,6 +14,7 @@ def save
     fix_image_param!
     fix_url!
     collection.attributes = params
+    collection.url_slug = CreateURLSlug.call(@slug) if params[:url_slug]
```
jgondron: @slug? I feel like a spec would have exposed this as a bug :trollface:

rfox2: It did - I just don't like being exposed. I'll fix it. :-)

rfox2: Corrected

app/services/save_collection.rb
```diff
@@ -14,7 +14,7 @@ def save
     fix_image_param!
     fix_url!
     collection.attributes = params
-    collection.url_slug = CreateURLSlug.call(@slug) if params[:url_slug]
+    collection.url_slug = CreateURLSlug.call(params[:url_slug]) if params[:url_slug]
```
jgondron: This needs a spec

jgondron:
👍

PR #372, opened 2016-06-08
DEC-1045 Create service to return collection via custom slug

rfox2:
What: Implemented a service that returns a collection api object based on the value of a custom URL slug that the user creates.
How: There's a new controller method that takes a slug as an argument and emits the collection object. A 404 will be returned if there is no collection matching the slug value. The controller method in turn calls the collection query class that has a new method to find a published collection based on the slug value. The collection model now also has a custom_slug() method to return the url_slug value the user creates. The JSON builder was modified to include the custom_slug value in the output if one is present.

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -33,8 +33,4 @@ class Collection < ActiveRecord::Base
   def item_json_url(item_id)
     "/api/collections/#{id}/items/#{item_id}.json?include=image"
   end
-
-  def custom_slug
-    url_slug
-  end
 end
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -33,4 +33,8 @@ class Collection < ActiveRecord::Base
   def item_json_url(item_id)
     "/api/collections/#{id}/items/#{item_id}.json?include=image"
   end
+
+  def custom_slug
+    url_slug
+  end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -68,6 +68,7 @@ Rails.application.routes.draw do
   resources :sections, only: [:edit, :update, :destroy]
 
   namespace :v1 do
+    get :custom_slug, path: "/collections/custom_slug/:slug", to: "collections#custom_slug", defaults: { format: :json }
     resources :collections,
               only: [:show, :index],
               defaults: { format: :json } do
--- a/spec/controllers/v1/collections_controller_spec.rb
+++ b/spec/controllers/v1/collections_controller_spec.rb
@@ -2,14 +2,14 @@ require "rails_helper"
 require "cache_spec_helper"
 
 RSpec.describe V1::CollectionsController, type: :controller do
-  let(:collection) { FactoryGirl.create(:collection, published: true, site_path: "[]", url_slug: "test") }
+  let(:collection) { instance_double(Collection, id: 1, published: true, site_path: "[]", url_slug: "test") }
   let(:collections) { [collection] }
 
   before(:each) do
     allow_any_instance_of(CollectionQuery).to receive(:public_collections).and_return(collections)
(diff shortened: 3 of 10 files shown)


danthewolfe:
👍

spec/controllers/v1/collections_controller_spec.rb
```diff
@@ -59,6 +60,30 @@
...
+      expect(subject).to render_template("v1/collections/show")
+    end
+
+    it_behaves_like "a private basic custom etag cacher"
+
+    it "uses the V1Collections#show to generate the cache key" do
+      expect_any_instance_of(CacheKeys::Custom::V1Collections).to receive(:show)
+      subject
+    end
+  end
```
jgondron: What about when there is no collection with that slug?

spec/queries/collection_query_spec.rb
```diff
@@ -49,6 +49,20 @@
     end
   end
 
+  describe "#custom_slug_find" do
+    it "returns collection with specific custom slug" do
+      expect(relation).to receive(:where).
+        with("url_slug = ? AND (published = ? OR preview_mode = ?)", "test_slug", true, true).
+        and_return(@collection_array = [collection])
+      expect(@collection_array).to receive(:take!).and_return(collection)
+      subject.custom_slug_find("test_slug")
```
jgondron: This test isn't ensuring that the method returns the correct value. This should likely be broken up into two tests, one that makes sure it uses the correct query, and another that tests the return value.

config/routes.rb
```diff
@@ -68,6 +68,7 @@
   resources :sections, only: [:edit, :update, :destroy]
 
   namespace :v1 do
+    get :custom_slug, path: "/collections/custom_slug/:slug", to: "collections#custom_slug", defaults: { format: :json }
```
jgondron: Can this be written as part of the collections resource?

app/models/collection.rb
```diff
@@ -33,4 +33,8 @@ def items_json_url
   def item_json_url(item_id)
     "/api/collections/#{id}/items/#{item_id}.json?include=image"
   end
+
+  def custom_slug
```
rfox2: I'd like to have a method that differentiates any other use of "slug" which we do have on the collection model. It's easier to distinguish in other contexts.

config/routes.rb
rfox2: No because it would get entangled in the rails "resourceful routes" mapper. Trust me I tried.

spec/controllers/v1/collections_controller_spec.rb
rfox2: A 404 will get sent back. I'll write a test for that.

spec/queries/collection_query_spec.rb
rfox2: I've added an additional test and also cleaned up the first one

app/models/collection.rb
rfox2: @jgondron and I talked about this and I'll just remove references to custom slug for now.

spec/controllers/v1/collections_controller_spec.rb
```diff
@@ -2,14 +2,14 @@
 require "cache_spec_helper"
 
 RSpec.describe V1::CollectionsController, type: :controller do
-  let(:collection) { instance_double(Collection, id: 1, published: false, site_path: "[]", url_slug: "test") }
+  let(:collection) { FactoryGirl.create(:collection, published: true, site_path: "[]", url_slug: "test") }
```
jgondron: Any reason to create an actual record here?

rfox2: I replaced it with a double.

jgondron:
👍

PR #373, opened 2016-06-09
DEC-1050: Metadata only icon

jgondron:
Changed the item listing to use the tag icon (https://design.google.com/icons/#ic_local_offer) when no image is present to signify that the item will only render metadata tags.


rfox2:
👍

danthewolfe:
This is a pretty big change. Maybe we should wait until Jon can give a 3rd thumbs up...

👍

jgondron:
@danthewolfe Just called Jon, he said 👍 . He also said not to call him again while he's 🍹 🌴

PR #374, opened 2016-06-10
DEC-1039: Custom facets

jgondron:
We can customize facets for any client by modifying the configuration, but currently we had no way to customize the limits that are used when displaying each facet. Changed the facet object to allow specifying a limit per facet. If none is found, it will fall back to the default in the solrconfig.xml


rfox2:
👍

danthewolfe:
👍

PR #375, opened 2016-06-10
DEC-1040: Custom sorts

jgondron:
Why: User would like to be able to specify custom sort options, including different directions for the same field.
How: Had to reinstate the config field lookups in the sort field and sort direction calls. This will allow specifying different directions for each sort option to use by default, ex: sort name asc, but date desc. This only comes into play when the caller does not specify a direction param. If a direction is not present in the params, or the config, it will fall back to "desc" to match the default solr behavior.

### DIFF
--- a/app/models/waggle/search/query.rb
+++ b/app/models/waggle/search/query.rb
@@ -24,18 +24,28 @@ module Waggle
 
       def sort_field
         if sort.present?
-          sort.split(" ")[0].to_sym
+          sort_sym = sort.split(" ")[0].to_sym
+          sort_field = configuration.sort(sort_sym)
+          return sort_field.field_name if sort_field.present?
+          sort_sym
         end
       end
 
       def sort_direction
         if sort.present?
+          dir = "desc"
           sort_arr = sort.split(" ")
-          return "desc" if sort_arr.length < 2
-          dir = sort_arr[1]
-          return "desc" if dir != "desc" && dir != "asc"
-          dir
+          if sort_arr.length < 2
+            sort_field = configuration.sort(sort_arr[0].to_sym)
+            return get_valid_sort_direction(sort_field.direction) if sort_field.present?
+          end
+          get_valid_sort_direction(sort_arr[1])
         end
+      end
+
+      def get_valid_sort_direction(direction)
+        return "desc" if direction != "desc" && direction != "asc"
+        direction
       end
 
       def result
--- a/app/models/waggle/search/query.rb
+++ b/app/models/waggle/search/query.rb
@@ -33,7 +33,6 @@ module Waggle
 
       def sort_direction
         if sort.present?
-          dir = "desc"
           sort_arr = sort.split(" ")
           if sort_arr.length < 2
             sort_field = configuration.sort(sort_arr[0].to_sym)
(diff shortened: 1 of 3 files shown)


danthewolfe:
👍

app/models/waggle/search/query.rb
```diff
@@ -24,20 +24,29 @@ def facet(name)
...
       end
 
       def sort_direction
         if sort.present?
           sort_arr = sort.split(" ")
-          return "desc" if sort_arr.length < 2
-          dir = sort_arr[1]
-          return "desc" if dir != "desc" && dir != "asc"
-          dir
+          if sort_arr.length < 2
```
rfox2: At this point, wouldn't the array value always be == 1? Just wondering if there's value in checking for 0.

app/models/waggle/search/query.rb
```diff
@@ -24,20 +24,29 @@ def facet(name)
 
       def sort_field
         if sort.present?
-          sort.split(" ")[0].to_sym
+          sort_sym = sort.split(" ")[0].to_sym
+          sort_field = configuration.sort(sort_sym)
+          return sort_field.field_name if sort_field.present?
```
rfox2: Why would the value returned from the configuration ever differ from what gets submitted in the sort array from the client?

app/models/waggle/search/query.rb
```diff
@@ -24,20 +24,29 @@ def facet(name)
...
+          if sort_arr.length < 2
+            sort_field = configuration.sort(sort_arr[0].to_sym)
+            return get_valid_sort_direction(sort_field.direction) if sort_field.present?
+          end
+          get_valid_sort_direction(sort_arr[1])
         end
       end
 
+      def get_valid_sort_direction(direction)
+        return "desc" if direction != "desc" && direction != "asc"
```
rfox2: I think the default should be "diag" all things being equal.

spec/models/waggle/search/query_spec.rb
```diff
@@ -53,24 +74,34 @@
...
     end
 
     it "defaults to desc if invalid second param is given for sort option" do
       arguments[:sort] = "field blah"
       expect(subject.sort_direction).to eq("desc")
     end
+
+    it "defaults to desc if there is a field name, but no direction is given in the arguments or the config" do
+      allow(sort).to receive(:direction).and_return("blah")
+      expect(subject.sort_direction).to eq("desc")
```
rfox2: Should be "diag". Just my 2 cents.

app/models/waggle/search/query.rb
jgondron: It's the same. At some point I probably felt like it reads easier this way, because you can see that args over 2 would fall through to the next line, instead of thinking about 0. I can change if it reads easier as == 1

app/models/waggle/search/query.rb
jgondron: The sorts in solr are a union of two sets: the fields defined in the configs, and a few other fields such as updated_at. If the field is not present in the config, it passes through to solr anyway in case that sort field was defined elsewhere in the schema.

app/models/waggle/search/query.rb
jgondron: How about "left"

rfox2:
👍

PR #376, opened 2016-06-15
Dec 1064 update item metadata saving

jonhartzler:
Why: the current implementation was validating the metadata on create which was causing new item creation to fail also there was a bug with the custom metadata validator in that it always used the first loaded collection configuration to validate agains.
What:  I separated the routes for updating items and metadata.  All routes to items should only save item data.  There is a new api route v1/items/ID/metadata that saves the metadata.   There is also a new save_metadata class for saving metadata.  
SaveItem no longer saves metadata.

### DIFF
--- a/app/services/save_metadata.rb
+++ b/app/services/save_metadata.rb
@@ -1,0 +1,33 @@
+class SaveMetadata
+  attr_reader :params, :item
+
+  def self.call(item, params)
+    new(item, params).save
+  end
+
+  def initialize(item, params)
+    @params = params
+    @item = item
+  end
+
+  def save
+    set_metadata
+    if valid? && item.save
+      item.item_metadata
+    else
+      false
+    end
+  end
+
+  private
+
+  def set_metadata
+    if params.present?
+      Metadata::Setter.call(item, params)
+    end
+  end
+
+  def valid?
+    item.item_metadata.valid?
+  end
+end
--- a/app/services/save_metadata.rb
+++ b/app/services/save_metadata.rb
@@ -28,9 +28,7 @@ class SaveMetadata
   end
 
   def set_metadata
-    if params.present?
-      Metadata::Setter.call(item, params)
-    end
+    Metadata::Setter.call(item, params)
   end
 
   def valid?
--- a/app/services/save_metadata.rb
+++ b/app/services/save_metadata.rb
@@ -11,6 +11,7 @@ class SaveMetadata
   end
 
   def save
+    fix_params
     set_metadata
     if valid? && item.save
       item.item_metadata
@@ -20,6 +21,11 @@ class SaveMetadata
   end
 
   private
+
+  def fix_params
+    @params = params.with_indifferent_access
+    ParamCleaner.call(hash: params)
+  end
 
   def set_metadata
     if params.present?
--- a/spec/controllers/v1/metadata_controller_spec.rb
+++ b/spec/controllers/v1/metadata_controller_spec.rb
@@ -2,10 +2,6 @@ require "rails_helper"
 require "cache_spec_helper"
 
 RSpec.describe V1::MetadataController, type: :controller do
-  let(:collection_configuration) { CollectionConfiguration.new }
-  let(:collection) { instance_double(Collection, id: "1", updated_at: nil, items: nil, collection_configuration: collection_configuration) }
-  let(:item) { instance_double(Item, id: "1", collection: collection, children: nil) }
-
   before(:each) do
     allow_any_instance_of(ItemQuery).to receive(:public_find).and_return(item)
     allow_any_instance_of(CollectionQuery).to receive(:public_find).and_return(collection)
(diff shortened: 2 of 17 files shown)


spec/services/save_metadata_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  it "sets the metadata with the metadata setter" do
+    expect(Metadata::Setter).to receive(:call).with(item, params).and_return(true)
+    subject
+  end
+
+  it "does not set the metadata if the params are nil" do
+    expect(Metadata::Setter).to_not receive(:call).with(params)
+    params = nil
```
jonhartzler: hound has this wrong because it does not know about the let statement ahead.  I am not planning on correcting this.

spec/services/save_metadata_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    allow(Metadata::Setter).to receive(:call).and_return(true)
+  end
+
+  it "sets the metadata with the metadata setter" do
+    expect(Metadata::Setter).to receive(:call).with(item, params).and_return(true)
+    subject
+  end
+
+  it "does not set the metadata if the params are nil" do
+    expect(Metadata::Setter).to_not receive(:call).with(params)
```
jgondron: Don't need the with(params)

spec/services/save_metadata_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    params = nil
+    subject
+  end
+
+  it "calls item#save if the item_metadata is valid? " do
+    expect(item_metadata).to receive(:valid?).and_return(true)
```
jgondron: Should just be an allow.

spec/services/save_metadata_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    expect(item).to receive(:save).and_return(true)
+    subject
+  end
+
+  it "does not call item#save if the item_metadata is not valid?" do
+    expect(item_metadata).to receive(:valid?).and_return(false)
```
jgondron: Should just be an allow.

spec/services/save_metadata_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    expect(item).to_not receive(:save)
+    subject
+  end
+
+  it "returns the item_metadata if the process is successful" do
+    expect(subject).to eq(item_metadata)
+  end
+
+  it "returns false if the the item_metadata is not valid?" do
+    expect(item_metadata).to receive(:valid?).and_return(false)
```
jgondron: Should just be an allow.

spec/services/save_metadata_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    expect(subject).to eq(false)
+  end
+
+  it "returns false if the the item save is false" do
+    expect(item).to receive(:save).and_return(false)
```
jgondron: Should just be an allow.

spec/controllers/v1/metadata_controller_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  let(:collection) { instance_double(Collection, id: "1", updated_at: nil, items: nil, collection_configuration: collection_configuration) }
+  let(:item) { instance_double(Item, id: "1", collection: collection, children: nil) }
+
+  before(:each) do
+    allow_any_instance_of(ItemQuery).to receive(:public_find).and_return(item)
+    allow_any_instance_of(CollectionQuery).to receive(:public_find).and_return(collection)
+  end
+
+  describe "PUT #update" do
+    let(:collection_configuration) { CollectionConfiguration.new }
```
rfox2: Why is this needed when you have :collection_configuration assigned above?

spec/controllers/v1/metadata_controller_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      expect(response).to be_unprocessable
+    end
+
+    it "renders with errors" do
+      allow(SaveMetadata).to receive(:call).and_return(false)
+      subject
+      expect(response).to render_template("errors")
+    end
+
+    it "assigns and item" do
```
rfox2: Should this be "assigns an item"?

app/services/save_metadata.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def initialize(item, params)
+    @params = params
+    @item = item
+  end
+
+  def save
+    fix_params
+    set_metadata
+    if valid? && item.save
```
rfox2: Should there be a spec to test for valid metadata? i.e. - a spec that tests that false gets returned if the metadata is invalid, and a spec that tests to see if the metadata is returned on success?

spec/controllers/v1/metadata_controller_spec.rb
jonhartzler: I removed the unused let statements from above.

jgondron:
👍

rfox2:
👍

PR #378, opened 2016-06-16
DEC-1065 Adios adios

rfox2:
What: Unfortunately, we have to say "adios" to "Adios", as much as I hate to.
How: Super secret complicated process that involves aliens, FBI agents, the NSA and former KGB members.


danthewolfe:
bueno

jgondron:
👍

PR #379, opened 2016-06-16
DEC-624: Bugfix item list on Safari

jgondron:
Why: Cannot see items for a collection in Safari
How: Problem was with material-ui Table. Symbol is undefined. We're pretty close to a deployment so I don't want to go through updating react and material-ui just to fix this, defined Symbol instead, see https://github.com/callemall/material-ui/issues/1531. Also fixed a problem with passing the key prefix to the pagination component. this.props.key was deprecated, need to pass as a different prop.

### DIFF
--- a/app/assets/javascripts/components.js
+++ b/app/assets/javascripts/components.js
@@ -1,6 +1,12 @@
 //app/assets/javascripts/components.js
 //= require_self
 //= require react_ujs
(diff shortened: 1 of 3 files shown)


rfox2:
👍

jonhartzler:
👍

PR #380, opened 2016-06-16
DEC-1068 Fix for double render bug

rfox2:
Bug: Rails was throwing a double render error randomly, possibly due to caching or jbuilder
Fix: In the end, I updated jbuilder, added a template for custom_slug, and rewrote some specs


jonhartzler:
👍

jgondron:
👍

PR #381, opened 2016-06-16
DEC-1069 Fix for custom slug update

rfox2:
Bug: Custom slug updates via the UI were failing
Fix: Thought that this was related to caching but it ended up being related to the permitted save params. This was due to changes in the model and removing the custom_slug method earlier. Adding url_slug to the permitted params list resolved the issue.


jgondron:
👍

jonhartzler:
👍

PR #382, opened 2016-06-17
DEC-1071 Fix broken preview link

rfox2:
Bug: Preview link was broken because of missing forward slash
Fix: I put the forward slash back in, but as a caveat, I previously needed to take the slash out because it was rendering a double forward slash - which broke the preview link as well


danthewolfe:
👍

rfox2:
👍  (from Jon)

PR #387, opened 2016-06-22
Dec 948 - update the publishing options for dec

jonhartzler:
1.  I removed the preview option.  All collections are viewable with a link at all times.  This is actually how it already was so it just needed to update the UI. 

2.  I updated the left nav to have a copy url feature same as zuchetto.  And a link to preview the site.  

3.  I updated the settings to have a publish options section.  This has UI to allow the collection to appear on collections.library.nd.edu.  I reused the published field for this.
screen shot 2016-06-22 at 10 42 10 am


danthewolfe:
:+1:

danthewolfe:
I like the copy link feature.

jonhartzler:
@danthewolfe i should hope so you wrote the original code!!

danthewolfe:
That's probably why I like it so much.

jgondron:
👍

PR #389, opened 2016-06-23
Configuration api updates

jonhartzler:
updates to the configuration so that it can put configuration options of enable search and browse in the configuration.


rfox2:
👍

jgondron:
Why was this needed?

danthewolfe:
👍

PR #391, opened 2016-06-27
DEC-1090 Fix for custom slug bug

rfox2:
What: CreateURLSlug was returning "name" when no value was supplied. This caused a bug for the custom slug functionality.
Fix: Changed the CreateURLSlug class so that it returns nil if no value is supplied, and modified the associated spec.


jgondron:
👍

rfox2:
👍

PR #393, opened 2016-06-28
DEC-841 Fix for creating metadata fields with same label / name

rfox2:
What: Implemented a fix for condition when curator creates a duplicate metadata field
Fix: Created a class that generates a uuid that will be assigned to every custom metadata field. This avoids the problem of having truly duplicate fields, but does not address the possibility that a curator will create two metadata fields with the same label which presents a problem for serializing collections from the API. That will need to be addressed in another ticket.


danthewolfe:
👍

jgondron:
👍

PR #394, opened 2016-06-29
DEC-1093 Prevent duplicate metadata label data entry

rfox2:
What: This is part 2 of the fix for entering a duplicate metadata field in the collection configuration. This PR addresses the potential problem of a curator entering in a duplicate metadata label.
Fix: Added a method to check for a duplicate label value against all of the extant metadata fields. It reports back with a true or false. create_field() in CreateConfigurationField now has a conditional to test for the duplicate label and returns nil if the entered parameter is a duplicate. The standard controller behavior was left untouched.


jgondron:
👍

danthewolfe:
👍

PR #395, opened 2016-07-01
DEC-1079: Refactor Item images

jgondron:
Why: In order to prepare for allowing an Item to have an associated image, video, or audio file, we need to refactor the way images are handled. This also means that we have to refactor the way the images are handled on Collections and Showcases as well.
How:
- Merged HoneypotImage with Image entity. For now we are still going to store the raw json_response from honeypot in Image. Future updates may change this to be something more inline with IIIF, ex: we store the base_url only.
- Migrated all of the paperclip data that was attached to Collection/Showcase objects into a new Image object and archived the old fields for now. Because of the complexity of this migration, I had to do a lot of this in the app layer, so the migration will take some time to perform (approx 1.5hrs on our prod server)
- Several service classes and decorators were written to handle multiple types of objects. These have been simplified to only handle an Image object.
- Honeycomb no longer stores any derivatives, only the original file. Because of this, we are going to regress slightly: 
  - You will not be able to use tiff files in development. This does not impact prod or preprod servers.
  - We will not do any limitation on pixel size (was doing 16m pixel max). The responsibility for limiting this will be pushed off to Honeypot. We will do this on Honeypot when it becomes a problem, if ever.


app/assets/javascripts/components/ItemShowImageBox.jsx
```diff
@@ -72,17 +72,17 @@ var ItemShowImageBox = React.createClass({
...
         return this.itemProcessingHtml();
       case "no_image":
         return this.itemNoImageHtml();
-      case "image_unavailable":
+      case "unavailable":
         return this.itemImageInvalidHtml();
       default:
-        EventEmitter.emit("MessageCenterDisplay", "error", "Unknown Image Status");
-        break;
+        console.log("Unknown Image Status: " + this.state.item.image_status);
```
rfox2: Did you mean to leave this statement in?

db/migrate/20160621142819_add_honeypot_fields_to_image.rb
```diff
@@ -0,0 +1,6 @@
+class AddHoneypotFieldsToImage < ActiveRecord::Migration
```
rfox2: How will this be used?

db/migrate/20160621151223_move_honeypot_images_to_image.rb
```diff
@@ -0,0 +1,103 @@
+class MoveHoneypotImagesToImage < ActiveRecord::Migration
```
rfox2: This is, by far, the biggest migration class I've ever seen.

rfox2:
Might want to look at some of the hound comments, and I had some minor questions but this looks good. 👍

app/assets/javascripts/components/ItemShowImageBox.jsx
jgondron: Yes. If this ever happens, it means there is some disconnect between the statuses that the api has defined and the way the front end is consuming those statuses, so theres nothing the user can do about it. I thought it would be better to just show them the standard error message, let them report the issue, then when we pull up the item we can diagnose on the console.

db/migrate/20160621142819_add_honeypot_fields_to_image.rb
jgondron: We decided we don't want to have two different types of images (ex: local images, and honeypot images). All images are now honeypot images. That means merging the Image and HoneypotImage entities. 

The json_response field will be used the same way we've been using it for other honeypot images. Honeypot will return the json for the image, we store it and render that via our api anytime a request is made for that image.

The status is a sort of FSM field for managing the lifecycle of an image. This used to be on item, but was really an FSM for the associated images that are sent to an external image server for processing. We need to know if the item is in processing, processed and ready, or has some problem with it so that we can handle those cases appropriately. Since all images are now honeypot images, keeping a status for all images makes more sense.

jgondron: Also note that this eventually means dropping the HoneypotImage entity in a future migration (once everything is known to be stable)

db/migrate/20160621151223_move_honeypot_images_to_image.rb
jgondron: Yeah, unfortunately, I had to do a lot of things in the Application layer, so it meant I had to ensure a deterministic state of a lot of the objects. Not as pretty as I would like, but hopefully more reliable.

jonhartzler:
👍

PR #397, opened 2016-07-06
Remove status column from Collections Home Page

laritakr:
Removed status column from Collections home page.

Deleted column from _list partial, and removed the entire PublishedText class which is now obsolete.


rfox2:
👍

jonhartzler:
👍

PR #398, opened 2016-07-06
Set homepage as active option on Site Setup page

laritakr:
Added homepage as the default active option upon entry to Site Setup page.

Added code into collections_side_nav.rb to set returned value to "active" for homepage as a default option. This is used in the _site_nav partial to set the class value.

### DIFF
--- a/app/decorators/collections_side_nav.rb
+++ b/app/decorators/collections_side_nav.rb
@@ -29,6 +29,7 @@ class CollectionsSideNav < Draper::Decorator
   def active_tab_class(tab:)
     return "active" if tab == form
     return "active" if tab == "edit" && form.nil?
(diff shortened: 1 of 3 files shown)


rfox2:
👍

jgondron:
👍

app/decorators/collections_side_nav.rb
```diff
@@ -29,6 +29,7 @@ def site_path_link
   def active_tab_class(tab:)
     return "active" if tab == form
     return "active" if tab == "edit" && form.nil?
```
jonhartzler: Should this line continue to exist ?   I don't think tab edit exists anymore.  
Current tabs are homepage|collection_introduction|about_text|copyright_text|site_path   so this should be removed.

PR #399, opened 2016-07-06
Dec 1102 remove name save and description save

laritakr:
Make a migration to remove: name_save: text, description_save: text from items.

Also bundle update hesburgh_infrastructure.


jonhartzler:
👍

jgondron:
👍

PR #400, opened 2016-07-06
Dec 1103 remove the ui and database field to create introductions

laritakr:
Migrate database to remove site_intro column from collections. Removed the collection introduction form and related routes, as well as all other references to site_intro.


jgondron:
We can review this, but let's hold off on merging just yet. We need to discuss when we are going to deploy next.

jgondron:
👍

jonhartzler:
👍

PR #404, opened 2016-07-07
DEC-1098 fix the drag and the scrolling.

jonhartzler:
As a side note.  There are several warnings I did not correct.  If some one believes I should go back I can try to address those.


rfox2:
👍

jgondron:
👍

PR #405, opened 2016-07-08
DEC-1101 Fix for preview link urls

rfox2:
Bug: The preview links on the item edit and showcase edit pages were broken when a custom slug was established. The preview links tried to use the custom url as the base url for the preview.
Fix: Added a flag in the method call for CreateBeehiveURL that will output the custom url if the flag is set to true; by default the flag is set to false, and it's an optional parameter for the call() method.


jgondron:
👍

danthewolfe:
:+1:

PR #407, opened 2016-07-12
Standardize page styles for editors and setup

laritakr:
Remove panels and add standard heading styles for Site Setup and Editors pages. 
Rename title on setup sidebar to be consistent with main menu option name.

Corresponds to JIRA tasks DEC-1138 and DEC-1139.


jonhartzler:
👍

PR #408, opened 2016-07-12
Fix deploy issue

jgondron:
Browserify was choking on the MetaDataConfigurationForm files. Problem was due to case sensitivity on the build filesystem that doesn't exist on OSX. A few files and directories were named Metadata instead of MetaData.


jonhartzler:
👍

PR #411, opened 2016-07-14
fix issue with redactor not saving

jonhartzler:
redactor updated the interface for how events are added.  this was not working in the react class.


jgondron:
👍

danthewolfe:
👍

PR #412, opened 2016-07-14
code climate

jonhartzler:
Fixed this so code climate does not show a critical warning.  I know this is smoke and mirrors...


rfox2:
👍

PR #413, opened 2016-07-14
Add super- and subscript buttons to redactor

laritakr:
Add new superscript and subscript options to text boxes.
Move redactor modules into separate folder for clarity.


jgondron:
👍

PR #415, opened 2016-07-14
Standardize format of item edit page

laritakr:
Add title to page. Remove panels from javascript form.

Also modified title for edit page to only say edit.


jonhartzler:
👍

PR #416, opened 2016-07-20
DEC-1148 Fix for missing name import bug

rfox2:
Bug: Import routine was raising exception when name was missing from import spreadsheet
Fix: Added check for the validity of item metadata since the name property was moved into the metadata attribute

### DIFF
--- a/app/services/create_items.rb
+++ b/app/services/create_items.rb
@@ -31,7 +31,7 @@ class CreateItems
         item_props = yield(item_props, rewrite_errors).symbolize_keys if block_given?
         item_creator = FindOrCreateItem.call(props: { collection_id: collection_id, **item_props }, find_by: find_by)
         saved = rewrite_errors.present? ? false : item_creator.save
-        add_to_errors(errors: errors, index: index, new_errors: rewrite_errors | item_creator.item.errors.full_messages, item: item_creator.item)
+        add_to_errors(errors: errors, index: index, new_errors: rewrite_errors | item_creator.item.errors.full_messages | item_creator.item.item_metadata.errors.full_messages, item: item_creator.item)
         update_counts(save_successful: saved, item: item_creator, counts: counts)
       end
     end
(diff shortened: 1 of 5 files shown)


app/services/create_items.rb
```diff
@@ -31,7 +31,12 @@ def create_or_update!(collection_id:, find_by:, items_hash:, counts:, errors:)
         item_props = yield(item_props, rewrite_errors).symbolize_keys if block_given?
         item_creator = FindOrCreateItem.call(props: { collection_id: collection_id, **item_props }, find_by: find_by)
         saved = rewrite_errors.present? ? false : item_creator.save
-        add_to_errors(errors: errors, index: index, new_errors: rewrite_errors | item_creator.item.errors.full_messages, item: item_creator.item)
+        add_to_errors(
+          errors: errors,
+          index: index,
+          new_errors: rewrite_errors | item_creator.item.errors.full_messages | item_creator.item.item_metadata.errors.full_messages,
```
jonhartzler: Did you check uploading an image still works with the drag and drop editor?  
I am concerned that if they are sharing code paths item_creator.item.item_metadata.errors.full_messages  will cause the regular upload to fail.

app/services/find_or_create_item.rb
```diff
@@ -66,6 +66,6 @@ def update_props
     end
     item.assign_attributes(props)
     CreateUniqueId.call(item)
-    @is_valid = item.valid?
+    @is_valid = item.valid? && item.item_metadata.valid?
```
jonhartzler: Same as the other comment.  I just want to be sure that this will not cause problems with uploading through the forms.

app/services/create_items.rb
rfox2: Drag and drop for item creation still works

app/services/find_or_create_item.rb
rfox2: Looks OK

jonhartzler:
👍

danthewolfe:
👍

PR #417, opened 2016-07-26
DEC-1164 Initial fix for image processing bug

rfox2:
Bug: Image processing status was not being determined properly due to status changes made when images were separated from items, leaving the item image box in a limbo state when an image was replaced.
Fix: I put the correct status strings into the ItemShowImageBox component and that resolved the issue.


danthewolfe:
:+1:

PR #419, opened 2016-07-28
DEC-1153: Add media to database

jgondron:
This changes the Image model into a more general Media model with a type and json data field to store the information about the image media. This will be used to also store other media types such as audio and video. This does not implement the actual Audio or Video models.


rfox2:
Once hound is satisfied, 👍

jgondron:
Don't kid yourself, hound will never be satisfied.

danthewolfe:
:+1:

PR #424, opened 2016-08-08
DEC-1156: Create audio video models

jgondron:
This adds the Audio and Video models. I ended up making our own version of enum that works with store_accessors now that we have 3 different entities each with their own state system. It works just like rails enum, except you can pass a store accessor so that the enum value will be stored in that store instead of on the base of the model.

### DIFF
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -2,4 +2,6 @@ class Media < ActiveRecord::Base
   belongs_to :collection
   has_many :items
   validates :collection, presence: true
+
+  has_paper_trail
 end
--- a/app/models/video.rb
+++ b/app/models/video.rb
@@ -4,6 +4,4 @@ class Video < Media
   extend StoreEnum
   store_enum :data, status: { allocated: 0, ready: 1 }
   store_accessor :data, :json_response
-
-  has_paper_trail
 end
--- a/app/models/video.rb
+++ b/app/models/video.rb
@@ -1,0 +1,9 @@
+require "store_enum"
+
+class Video < Media
+  extend StoreEnum
+  store_enum :data, status: { allocated: 0, ready: 1 }
+  store_accessor :data, :json_response
+
+  has_paper_trail
+end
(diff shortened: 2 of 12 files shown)


jgondron:
Please ignore the hound issues on the store enum file. This was mostly a copy/paste of rails enum and I'd rather not drift too far from the original just to fix style guide issues.

app/models/video.rb
```diff
@@ -0,0 +1,7 @@
+require "store_enum"
+
+class Video < Media
+  extend StoreEnum
+  store_enum :data, status: { allocated: 0, ready: 1 }
+  store_accessor :data, :json_response
```
rfox2: This will contain the thumbnail URI - just an fyi

lib/store_enum.rb
```diff
@@ -0,0 +1,197 @@
+require 'active_support/core_ext/object/deep_dup'
+
```
rfox2: If we're going to assume this library into our code base, I'd recommend bringing it up to our style standards.

lib/store_enum.rb
```diff
@@ -0,0 +1,197 @@
...
+# Declare an enum attribute where the values map to integers in the database,
```
rfox2: Can these comments be abbreviated?

rfox2:
I didn't see that the enum code was a copy/paste from rails. I was suggesting we bring that up to our code standards if we're adopting that code into our base - most are pretty easy fixes. Apart from that, after the other Hound issues are fixed than 👍

app/models/media.rb
```diff
@@ -2,4 +2,6 @@ class Media < ActiveRecord::Base
   belongs_to :collection
   has_many :items
   validates :collection, presence: true
+
+  has_paper_trail
```
jonhartzler: Is there a spec for this ?

jgondron: I have a "has a papertrail" inside of media_spec.rb

jonhartzler:
👍

PR #425, opened 2016-08-08
Dec 1171

jonhartzler:
Added simple routing for honeycomb endpoints.  Added a some react classes to start the process of creating 2 forms for updating different media.


app/assets/javascripts/actions/ItemActions.js
```diff
@@ -88,6 +88,19 @@ class ItemActions extends NodeEventEmitter {
   url(id) {
     return "/v1/items/" + id;
   }
+
+  getSignedUrl(id) {
```
jgondron: I'd be careful about adding actions that don't interact with a store. This is just adding extra complexity where it may not be needed. If this piece of data needed to be shared between components, then make an action and have that action populate the store, then store would broadcast a change. In this case, it's only one component that cares about this.

jgondron:
I'm ignoring most of the backend code and lack of specs for now. I'll add that as part of DEC-1173. If you can address my other comment on the js, 👍

PR #426, opened 2016-08-10
DEC-1156: Create audio video media

jgondron:
This adds the API endpoint and all service classes for creating new video and audio. Also fixed a bug with reading the current value of a store_enum.


jonhartzler:
👍

PR #429, opened 2016-08-17
Dec 1174  Action to finish uploading an image

jonhartzler:
This PR does not cover post back to buzz if we decide to do the post back.   It only covers moving the state from allocated to ready.  We will need to discuss if we wish for post back to happen.  DEC-1175 references dealing with the buzz side of the transaction.  

Added an action to finish upload at 
/v1/media/#id/finish_upload. 

Added service class to perform the transition
Added Query object for media.

### DIFF
--- a/app/controllers/v1/media_controller.rb
+++ b/app/controllers/v1/media_controller.rb
@@ -14,10 +14,10 @@ module V1
     def finish_upload
       @media = MediaQuery.new.public_find
 
-      media = FinishMediaUpload.call(media: @media, params: finish_params)
+      FinishMediaUpload.call(media: @media)
 
-      status = media.valid? ? :ok : :unprocessable_entity
-      render json: media.to_json, status: status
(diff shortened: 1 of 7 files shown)


app/controllers/v1/media_controller.rb
```diff
@@ -11,6 +11,15 @@ def create_for_item
       render json: media.to_json, status: status
     end
 
+    def finish_upload
+      @media = MediaQuery.new.public_find
```
rfox2: Why do we still have these methods like "public_find"? Didn't we implement those when we were considering the notion of "published" vs. "not published"?

rfox2:
Just one question, but 👍

app/controllers/v1/media_controller.rb
jonhartzler: We did right now it is a question of consistency.   The changing of that method name should be PR that fixes them all.

danthewolfe:
👍

PR #430, opened 2016-08-18
DEC-1176 Add controller action for uploading media thumbnails

rfox2:
**Goal**: Provide a controller action that can be used to upload a media thumbnail image
**Work**: Added the method plus specs, and route

### DIFF
--- a/app/controllers/v1/media_controller.rb
+++ b/app/controllers/v1/media_controller.rb
@@ -20,10 +20,27 @@ module V1
       render json: @media.to_json, status: status
     end
 
+    def add_thumbnail
+      @media = MediaQuery.new.public_find(media_params[:id])
+      @uploaded_image = media_params[:uploaded_image]
+      return_value = SaveMediaThumbnail.call(
+        image: media_params[:uploaded_image],
+        item: @media.items[0],
+        media: @media
+      ) if @uploaded_image.present?
+
+      status = return_value ? :ok : :unprocessable_entity
+      render json: return_value.to_json, status: status
+    end
+
     private
 
     def create_params
       params.require(:medium)
+    end
+
+    def media_params
+      params[:media].to_hash.with_indifferent_access
     end
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/v1/media_controller.rb
```diff
@@ -20,10 +20,27 @@ def finish_upload
...
+      @media = MediaQuery.new.public_find(media_params[:id])
+      @uploaded_image = media_params[:uploaded_image]
+      return_value = SaveMediaThumbnail.call(
+        image: media_params[:uploaded_image],
+        item: @media.items[0],
+        media: @media
+      ) if @uploaded_image.present?
+
+      status = return_value ? :ok : :unprocessable_entity
+      render json: return_value.to_json, status: status
```
jgondron: Is this only rendering the json response from Honeypot? Or is it/should it be rendering the media object as a whole (using SerializeMedia)?

rfox2: Done

config/routes.rb
```diff
@@ -103,6 +103,7 @@
 
     resources :media, only: [] do
       put "finish_upload", to: "media#finish_upload"
+      put "add_thumbnail", to: "media#add_thumbnail"
```
rfox2: Done

app/services/serialize_media.rb
```diff
@@ -4,7 +4,8 @@ def self.to_hash(media:)
     when "Image"
       SerializeImageMedia.to_hash(media: media)
     else
-      SerializeAVMedia.to_hash(media: media)
+      result = SerializeAVMedia.to_hash(media: media)
+      result
```
jgondron: Is the result assignment needed? Seems like this is some left over debug code

rfox2: Ugh, yes it is

jgondron:
👍

jonhartzler:
👍

PR #432, opened 2016-08-22
Post to Buzz

jgondron:
Adds an action that will create the media object on Buzz. This will typically be called once the user has finished uploading to the designated s3 location. To do this I added a service class that will ideally handle all of the communication to the Buzz server and refactored the save_media_thumbnail service to use this instead of directly handling this communication.

### DIFF
--- a/app/services/save_media_thumbnail.rb
+++ b/app/services/save_media_thumbnail.rb
@@ -15,7 +15,7 @@ class SaveMediaThumbnail
     @image_response = send_image_request
     return false unless @image_response
     if update_media_record
-      @media_response.body
+      true
     else
       false
     end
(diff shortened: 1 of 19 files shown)


rfox2:
👍

app/services/save_media_thumbnail.rb
```diff
@@ -32,19 +32,11 @@ def send_image_request
...
-    if response.success?
-      response
-    else
-      false
-    end
-  end
-
   def update_media_record
-    @media_response = send_media_request
+    @media.thumbnail_url = @image_response.body["contentUrl"]
```
jonhartzler: Isn't content URL the full size image?

jonhartzler:
One question then 👍

app/services/save_media_thumbnail.rb
jgondron: Yes, this was done as part of this change: https://github.com/ndlib/honeycomb/commit/f2bc0893cd1514f1f55582b74a81519065e9b7be#diff-771f53e3b0e5494ce9cb7ef68d923165R83. I kept that same behavior when refactoring. Did you or Rob discuss using the full image?

PR #434, opened 2016-08-23
Post to Buzz

jgondron:
Buzz expects file name only, not the full url, when creating new media. Changed this in the create action.


jonhartzler:
👍

PR #436, opened 2016-08-24
[WIP] Media in pages

jgondron:
Image insert function on redactor now works with media. To do this we had to change the way the api renders honeypot images to include a thumbnailUrl so that we can treat all media in a consistent manner when grabbing its thumbnail. I also added a set of default small/medium thumbnails to use for each media type. Images do not need one since if an image exists, it will have appropriate thumbnail derivatives.

Note: Flagging this as WIP since it is based on branch DEC-1171. Should wait until that's merged in


app/assets/javascripts/components/forms/ItemForm/ReplaceMediaForm.jsx
```diff
@@ -10,53 +11,62 @@ var ReplaceMedia = React.createClass({
...
-          completeCallback={this.completeCallback}
-          formUrl={ ItemActions.url(this.props.item.id) }
-          initializeCallback={this.dropzoneInitialized}
-          method="put"
-          multifileUpload={ true }
-          paramName="item[uploaded_image]"
-        />
-    </div>
-    );
+    console.log(this.props.item);
```
rfox2: Did you mean to leave this in?

app/assets/javascripts/components/forms/ItemForm/StreamingForm.jsx
```diff
@@ -0,0 +1,134 @@
...
+    }
+  },
+
+  setItem(data) {
+    if (this.state.processing) {
+      this.setState({ item: data }, this.loadSigningUrl);
+    }
+  },
+
+  uploady: function() {
```
rfox2: Just curious: why "uploady"?

rfox2:
👍  when DEC-1171 is merged

jonhartzler:
👍

PR #438, opened 2016-08-25
Media in showcases

jgondron:
Modified the showcase editor to allow inserting audio and video items into a showcase. 

Note: The showcase editor still hits an old version of the api, so there was a bit of refactoring to get this to work. It now does gets from v1, but still uses the old api for create/update/delete.


jonhartzler:
👍

rfox2:
👍

PR #440, opened 2016-08-29
Read credentials from aws config

jgondron:
Changed the s3 url signing to read the credentials from the aws prescribed shared location and moved remaining settings out of secrets.


jonhartzler:
👍

PR #441, opened 2016-08-30
DEC-1177 Add media type to index and reindex of item

rfox2:
We needed to have an item reindexed when a thumbnail was added so that it would appear in the item API output, and the media type is now output in search results. Some of the index classes were touched, and the item reindex call was added to the media thumbnail class.

### DIFF
--- a/app/models/waggle/adapters/solr/search/hit.rb
+++ b/app/models/waggle/adapters/solr/search/hit.rb
@@ -46,6 +46,10 @@ module Waggle
             fetch_text(:description)
           end
 
+          def media
+            eval(fetch_string(:media))
+          end
+
           def creator
             fetch_text(:creator)
           end
--- a/app/models/waggle/adapters/solr/search/hit.rb
+++ b/app/models/waggle/adapters/solr/search/hit.rb
@@ -46,10 +46,6 @@ module Waggle
             fetch_text(:description)
           end
 
-          def media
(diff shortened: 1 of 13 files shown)


app/models/waggle/adapters/solr/search/hit.rb
```diff
@@ -46,6 +46,10 @@ def description
             fetch_text(:description)
           end
 
+          def media
+            eval(fetch_string(:media))
```
danthewolfe: Sounds risky.

danthewolfe:
:+1:

app/models/waggle/adapters/solr/search/hit.rb
jgondron: Is the use of eval needed? We should also handle this somewhere else, such as whatever class is responsible for mapping a Honeycomb item to a Waggle item.

app/models/waggle/item.rb
```diff
@@ -29,7 +28,11 @@ def collection_id
     end
 
     def type
-      TYPE
+      if media
+        media["@type"]
+      else
+        "Unknown"
```
jgondron: I think Jon mentioned actually having a "Metadata only" type item in this case. Idk what value we use, but something along those lines. Maybe we just fall back to "Item"?

jgondron:
👍  Once the metadata only type is addressed.

PR #442, opened 2016-09-12
update to the ruby version

jonhartzler:
The basic problem I was having was related to newer versions of gems requiring a higher version of ruby.  (min 2.2.5)  The problem is we will want to get honeycomb to be updated on pre prod and prod...


rfox2:
👍

jonhartzler:
This branch is a bit of a notification.  I have started conversations with Tom about getting the update to happen.

hbeachey:
👍

PR #444, opened 2016-09-13
add SectionsController 'new' and fix tests around createing with item id

hbeachey:


app/controllers/sections_controller.rb
```diff
@@ -1,12 +1,21 @@
...
+
   def create
     check_user_edits!(showcase.collection)
     @section = SectionQuery.new(showcase.sections).build
 
     # This is to translate from public item.unique_id to internal item.id
     # Once we convert all of this to use the v1 api, we won't need to translate
-    item = ItemQuery.new.public_find(params[:section].delete(:item_id))
-    @section.item = item if item
+    id = params[:section].delete(:item_id)
```
jonhartzler: One little thing might be to encode this logic in a named method.  
like item_id_from_params.  Something that describes what it is doing.   The idea being it is easier to read later and encodes purpose.   

That method could use ||=  and you would not need to do an assignment then a check afterwords.  

`
if item_id_from_params
  item = ItemQuery.new.public_find(item_id_from_params)
`

mostly I wanted to use the new review system.

PR #446, opened 2016-09-15
Refactor so all thumbnails go through one codepath

hbeachey:
Made thumbnail creation more generic - allowing for custom styling and different defaults
Renamed HoneypotThumbnail to Thumbnail
Added thumbnails to page list view


app/assets/javascripts/components/Thumbnail.jsx
```diff
@@ -4,17 +4,36 @@ var classNames = require("classnames");
...
-        <img src={this.props.thumbnailUrl} className="hc-thumbnail-image" />
-      );
-    } else {
-      return (<mui.FontIcon  className="material-icons">local_offer</mui.FontIcon>);
+    var thumbUrl = this.props.thumbnailUrl
+
+    if (!thumbUrl || thumbUrl == "") {
+      switch (this.props.thumbType) {
+        case 'item':
+          thumbUrl = "https://image.freepik.com/free-icon/black-paper-tag_318-33822.jpg";
```
rfox2: Just a question: do we want to hard code placeholder images like this?

hbeachey: This is the same question I had, Jon said just to keep it until we get real images. These are just arbitrary images from google, so if there's a better - more correct - way to do this I'm happy to change it.

jgondron:
Get with Andy on the styling/color of the icons. The previous icons were a little more on the grey side.

app/views/collections/_homepage_form.html.erb
```diff
@@ -2,7 +2,7 @@
 
 <%= f.input :short_intro, input_html: { class: 'honeycomb_redactor'  } %>
 
-<%= HoneypotThumbnail.display(f.object.image) %>
+<%= Thumbnail.display(f.object.image, "item", { maxWidth: "85px", maxHeight: "85px" }) %>
```
jgondron: Doesn't seem like this should use an item thumbnail. This is representing a collection's home page and this causes it to look like an item metadata tag.

app/assets/javascripts/components/Thumbnail.jsx
hbeachey: We have real icons now!

app/views/collections/_homepage_form.html.erb
hbeachey: Fixed, and tinted the icons to match the outline of meta-only-image which is already grey.

app/assets/javascripts/components/Thumbnail.jsx
rfox2: Awesome

PR #447, opened 2016-09-16
Selected tags sometimes were not returned

hbeachey:
DEC-1133
Tags selected in the UI were sometimes not returned in the results, and thus could not be rendered in the UI - making it impossible to deselect them.

I added a new query that _only_ retrieves these tag values, no results, and appends the tags to the tag list if not already present.


jgondron:
I'm ok with this for now. I think long term we need to start handling this type of problem on the front-end instead. We can create a hard limit to prevent extreme cases, but otherwise the front end should be responsible for deciding how to deal with large counts of facet values. So if we have time in this sprint, I'd like to examine that. If not, 👍

PR #448, opened 2016-09-16
Dec 1189

jonhartzler:
This pull request will remove items that are created when the media creation process fails for any reason if the process has already created an item.


app/assets/javascripts/actions/ItemActions.js
```diff
@@ -90,7 +90,7 @@ class ItemActions extends NodeEventEmitter {
     });
   }
 
-  delete(id) {
+  delete(id, message=true) {
```
hbeachey: np. To be clearer this could be 'emitMessage' currently it reads as the message to be emitted.

PR #449, opened 2016-09-19
Dec 1206

jonhartzler:
The problem was that items were not getting indexed after media was being uploaded. 

This corrects that.


app/services/create_media.rb
```diff
@@ -40,9 +40,14 @@ def create_and_associate(type:)
...
       owner.save
+      index_item
     else
       media.serializer = SerializeMedia
     end
     media
   end
+
+  def index_item
+    Index::Item.index!(owner) if owner.is_a?(Item)
```
jgondron: Not a big fan this. Something prior to this in the call stack should know that the associated object is an item and index it.

jonhartzler: I agree that this is buried.  But what do you mean by earlier?  The only earlier is the calling method save and that is only one level higher in the stack and that will increase the branch conditions on an already large method.   I think that there are 2 possible issues here that cause this.  1.  we have a premature optimization regarding owner instead of media.item.  2.  The other thing we can do is look to break more of the action orientation of the save higher up the call stack.  So you might make a object just for each action.  Then we would call CreateMedia and IndexItem from inside that space and not need the guard clause.

jgondron: Right, I'm looking more at option 2. I just don't want to get this into the space that we have with SaveHoneypotImage where it's doing more than it should and there's not a clear of a separation of concerns.

jonhartzler: Ok I agree with that let me see what I can come up with.  Should not take too long.

PR #451, opened 2016-09-22
Item deletion had bad links for associated pages/showcases

hbeachey:
The @id field in now links to the api pages instead of to the honeycomb UI. 

I've changed all pages and showcases to use the unique_id field for their uri instead so that we can easily parse the v1 url to link to these resources.

This is already getting very messy and I'm worried about the complexity and readability as we move forward. There's already discussion about how things should have been done - separating the API from the UI - but I think it's worth talking about seriously so this codebase can be more maintainable.

### DIFF
--- a/app/assets/javascripts/components/PagesPanel.jsx
+++ b/app/assets/javascripts/components/PagesPanel.jsx
@@ -61,9 +61,17 @@ var PagesPanel = React.createClass({
 
     return this.state.pages.map(function(page, index) {
       key = "page-" + page.id;
+      reg = new RegExp( '^(.*/)v1/(.*)$', 'i' );
+      strings = reg.exec(page["@id"]);
+
+      url = strings[0];
+      if(strings && strings.length == 3) {
+        url = strings[1] + strings[2] + "/edit";
+      }
+
       return (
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/components/PagesPanel.jsx
```diff
@@ -61,9 +61,17 @@ var PagesPanel = React.createClass({
 
     return this.state.pages.map(function(page, index) {
       key = "page-" + page.id;
+      reg = new RegExp( '^(.*/)v1/(.*)$', 'i' );
+      strings = reg.exec(page["@id"]);
```
jgondron: I'm fine with this if it's working. Just fyi though, there is an @id which points to the purl for the object, and there is an "id" that gives you the unique_id alone, so you could have just pulled it from there instead of using regex to extract the id from @id.

hbeachey: I went with this method for two reasons: 1) most places are parsing in this fashion elsewhere in the codebase and 2) it felt incorrect to hard-code how the final uri should look in case it changed slightly in the future (eg. `/path/to/[id]/edit`).

PR #452, opened 2016-09-23
DEC-1212 Huge thumbnails on IE

hbeachey:
List/Index pages now all go through one react component to generate their internals - just like the item list minus the search functionality. This should fix incorrect sized thumbnails.


rfox2:
Can we get Hound fixed and get this merged in?

PR #453, opened 2016-10-17
Dec 1232

hbeachey:
The item delete button didn't grey out when it was un-deletable. It now does :)


app/decorators/item_decorator.rb
```diff
@@ -120,4 +121,8 @@ def children_query_recent
   def children_query
     @children_query ||= ItemQuery.new(object.children)
   end
+
+  def can_delete?
```
jgondron: Double check this, but I think there is an expectation that we always call this with an ItemQuery object. If so, ItemQuery has a can_delete? method already implemented, so you would just need to delegate this call to the object.

PR #454, opened 2016-10-19
DEC-449

hbeachey:
Showcase images in the list page were not constrained in height.


hbeachey:
I don't think this PR will be needed once DEC-1212 is merged in.

PR #455, opened 2016-10-21
DEC-946

hbeachey:
Metadata delete icons were misplaced in safari, this kicks them back up into position.

### DIFF
--- a/app/assets/javascripts/components/forms/MetaDataConfigurationForm/MetaDataConfigurationListItem.jsx
+++ b/app/assets/javascripts/components/forms/MetaDataConfigurationForm/MetaDataConfigurationListItem.jsx
@@ -43,7 +43,7 @@ var MetaDataConfigurationListItem = React.createClass({
           tooltipPosition="top-center"
           onTouchTap={function() { this.props.handleRightClick(field.name) }.bind(this) }
         >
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/components/forms/MetaDataConfigurationForm/MetaDataConfigurationListItem.jsx
```diff
@@ -43,7 +43,7 @@ var MetaDataConfigurationListItem = React.createClass({
           tooltipPosition="top-center"
           onTouchTap={function() { this.props.handleRightClick(field.name) }.bind(this) }
         >
-          field.active && <FontIcon className="material-icons" color={Colors.grey500} hoverColor={Colors.red500}>{icon}</FontIcon>
+          <FontIcon className="material-icons" color={Colors.grey500} hoverColor={Colors.red500}>{icon}</FontIcon> field.active &&
```
jgondron: Is field.active logic even needed here? This is already within the if(field.active) block, so it will be true.

hbeachey: Turns out no one knows what's going on here. Gonna remove it.

hbeachey: This is fixed

PR #459, opened 2016-10-27
Users cannot replace an image that failed processing

jgondron:
There were a few things causing problems related to this:
1. If the image already existed in honeycomb, it was not kicking off a job to process it even if they reuploaded it. To fix, I changed the FindOrCreateImage service to always process the image, even when it already exists.
2. If honeypot threw an exception and rendered a 500, sneakers workers would treat it as a successful job and move it to done. To fix, I changed our handling of responses from Faraday requests and changed it to raise an exception for anything other than a 200 so that the worker would catch it and requeue the job. I also changed all other services that use Faraday to use this new pattern.
3. The item edit controller was not considering media when rendering the etag. Changed the cache key generators to include the media object.

I would have preferred to only reprocess the image when there were no remaining jobs left for that image, but I could not find a reliable way to do so. Because of this we may encounter some edge cases. One that I can think of is that if Honeypot dies, the user then replaces the image multiple times, when Honeypot returns it will attempt to replace the image multiple times for that item. This could get weird if they choose different images to replace it with during Honeypots downtime, but I can't imagine that would ever happen.


app/errors/buzz/raise_faraday_exception.rb
```diff
@@ -0,0 +1,15 @@
+require 'faraday'
+require File.expand_path('../errors', __FILE__)
+
+module Buzz
+  class RaiseFaradayException < Faraday::Response::Middleware
+    def on_complete(response)
+      case response[:status]
```
hbeachey: Won't this cause all 200's to raise InternalServerError? Or does `f.use [whatever]` only use it on responses other than 200.

spec/controllers/v1/items_controller_spec.rb
```diff
@@ -4,7 +4,7 @@
 RSpec.describe V1::ItemsController, type: :controller do
   let(:collection_configuration) { CollectionConfiguration.new }
   let(:collection) { instance_double(Collection, id: "1", updated_at: nil, items: nil, collection_configuration: collection_configuration) }
-  let(:item) { instance_double(Item, id: "1", collection: collection, children: nil) }
+  let(:item) { instance_double(Item, id: "1", collection: collection, children: nil, media: nil) }
```
hbeachey: I made these same changes in the thumbnail PR, not an issue, just some merge conflicts coming our way.

jgondron: Yeah I saw that. I'll deal with it when it happens

app/errors/buzz/raise_faraday_exception.rb
jgondron: It will, I meant to go back and review the buzz changes. Good catch, thanks

rfox2:
Is there any way we can fix the Hound error? I think we're missing out on the style checks without it.

app/services/save_honeypot_image.rb
```diff
@@ -52,6 +52,7 @@ def send_request
 
   def connection
     @connection ||= Faraday.new(image_server_url) do |f|
+      f.use Honeypot::RaiseFaradayException
```
jgondron: Injects a handler for each request, not just for exceptions. I named it as such so that it's clear what that handler is doing: just checking the response in order to raise an exception.

app/services/buzz_media.rb
```diff
@@ -17,20 +17,12 @@ def self.call_update(media:)
...
-      response.body["object"]
-    else
-      raise_exception(faraday_response: response)
-    end
+    response.body["object"]
   end
 
   def update
     response = media_server_connection.put("/v1/media_files/" + @media.json_response["@id"], update_json)
-    if response.success?
```
jgondron: Yes, but the problem is that this isn't bubbling up to the worker. The workers rely on exceptions being thrown in order to requeue the job

app/errors/buzz/raise_faraday_exception.rb
```diff
@@ -0,0 +1,15 @@
+require 'faraday'
+
+module Buzz
+  class RaiseFaradayException < Faraday::Response::Middleware
+    def on_complete(response)
+      case response[:status]
+      when 200
+      when 422
+        raise Buzz::UnprocessableEntity, response.body
```
jgondron: buzz_media_spec has some coverage of this. I was having some trouble getting a new spec directory working for errors, but I can spend some more time on that if needed.

app/errors/buzz/raise_faraday_exception.rb
```diff
@@ -0,0 +1,15 @@
+require 'faraday'
+
+module Buzz
+  class RaiseFaradayException < Faraday::Response::Middleware
+    def on_complete(response)
+      case response[:status]
+      when 200
```
jgondron: Possibly, I was primarily trying to recreate our previous logic here though, which only cared about these.

app/errors/buzz/raise_faraday_exception.rb
jgondron: This is fixed

app/errors/buzz/raise_faraday_exception.rb
rfox2: That's totally up to you. If the coverage is important for regression and documentation, then go for it. But I'm not pressing for it.

PR #464, opened 2016-11-10
Showcase editor improvements

hbeachey:
* Added dynamic search to showcase editor items
* Added nametags to items
* Updated material-ui 13.0->13.4 which required react 14.7->14.8

_this may require everyone to run `bundle exec rake tmp:clear` for client up update correctly_


app/assets/javascripts/components/showcase_editor/ItemList.jsx
```diff
@@ -1,5 +1,34 @@
...
+
+  componentWillReceiveProps: function(nextProps) {
+    if(this.props.items.length > 0) {
+      return;
+    }
+
+    var options = {}
+    for(var i = 0; i < nextProps.items.length; ++i) {
+      lowerKey = nextProps.items[i].name.toLowerCase();
+      options[lowerKey] = true;
```
jgondron: Items can have the same name. Using item name as a key will be problematic.

jgondron: Ah ok, I see now. I was assuming we were using search api to get these items, didn't see that you were doing this to perform a client side filter. We should really discuss this. If this is part of DEC-1299, I think you're still going to have significant problems with large collections. 

Btw name your branches based on the ticket id. Ex: DEC-1299-Showcase-editor-improvements. That way Jira will see it and associate it. Traceability++

PR #465, opened 2016-11-11
DEC-450 Quick delete button(s)

hbeachey:
The request was for a quick way to delete items.

Initial implementation was with checkboxes in the item listing view which would then mass-delete all selected items. Unfortunately, getting information for each item on weather or not it can be deleted was expensive and slow.

Instead, I went with a delete button on the list view for each item which will save a few page transitions at least. I added these buttons to the items/pages/and showcases listings.

### DIFF
--- a/app/assets/javascripts/components/DeleteButton.jsx
+++ b/app/assets/javascripts/components/DeleteButton.jsx
@@ -6,8 +6,6 @@ var ItemStore = require("../stores/ItemStore");
 var ItemActions = require('../actions/ItemActions');
 
 var DeleteButton = React.createClass({
-  mixins: [DialogMixin],
-
   propTypes: {
     type: React.PropTypes.oneOf(['item', 'showcase', 'page']).isRequired,
     id: React.PropTypes.string.isRequired,
@@ -97,6 +95,36 @@ var DeleteButton = React.createClass({
     } else {
       return([this.cancelDismiss(), this.dialogConfirm()]);
     }
+  },
+
+  cancelDismiss: function() {
+    return (
+        <mui.FlatButton
+        label="Cacnel"
+        primary={true}
+        onTouchTap={this.dismissMessage}
+      />
+    );
+  },
+
+  okDismiss: function() {
+    return (
+      <mui.FlatButton
+        label="OK"
+        primary={true}
+        onTouchTap={this.dismissMessage}
+      />
+    );
+  },
+
+  dialogConfirm: function() {
+    return (
+      <mui.FlatButton
+        label="Yes"
+        primary={true}
+        onTouchTap={this.confirmMessage}
+      />
+    );
   },
 
   render: function() {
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/components/EntriesList.jsx
```diff
@@ -52,15 +59,18 @@ var EntriesList = React.createClass({
...
 
   getDefaultProps: function() {
     return { hasEntryCount: false };
   },
 
-  openItem: function(rowNumber) {
-    var selectedId = this.props.entries[rowNumber].id;
-    window.location = this.props.openUrl.replace("<id>", selectedId);
+  openItem: function(rowNumber, columnId) {
+    if(columnId != 4) {
```
jgondron: Is there no way to have your delete callback consume the event and prevent it from propagating to the table cell? I'm sure there is a cleaner way to do this. At the very least, get rid of the magic number so that it's clear what this logic is doing.

app/assets/javascripts/components/search/SearchPage.jsx
```diff
@@ -105,11 +107,22 @@ var SearchPage = React.createClass({
   },
 
   openItem: function(rowNumber, columnId) {
-    var selectedId = SearchStore.hits[rowNumber]["@id"];
-    var reg = new RegExp( '^.*\/(.*)$', 'i' );
-    var string = reg.exec(selectedId);
-    var itemId = string[1];
-    window.location = "/items/" + itemId + "/edit";
+    if(columnId != 4) {
```
jgondron: Same comment as before.

app/assets/javascripts/components/DeleteButton.jsx
```diff
@@ -0,0 +1,132 @@
+'use strict'
+var React = require('react');
+var mui = require("material-ui");
+
+var ItemStore = require("../stores/ItemStore");
+var ItemActions = require('../actions/ItemActions');
+
+var DeleteButton = React.createClass({
+  mixins: [DialogMixin],
```
jgondron: We've been trying to move away from using mixins. Migrate the methods into this object, or make a generic dialog that you can use to handle this.

app/controllers/v1/pages_controller.rb
```diff
@@ -19,5 +19,16 @@ def show
                                            page: @page)
       fresh_when(etag: cache_key.generate)
     end
+
+    def destroy
```
jgondron: Remove this method from the old controller and route if it's no longer handling it.

app/controllers/v1/showcases_controller.rb
```diff
@@ -19,5 +19,16 @@ def show
                                            showcase: @showcase)
       fresh_when(etag: cache_key.generate)
     end
+
+    def destroy
```
jgondron: Remove this method from the old controller and route if it's no longer handling it.

app/controllers/v1/pages_controller.rb
hbeachey: Pages and showcases are still using these methods for the detail view delete buttons. I didn't change them over because the user experience is quite different and it makes sense (to me at least) to keep these the same. Perhaps that's not the case and I should re-write the other buttons as well.

PR #467, opened 2016-11-15
Import CSV

jgondron:
- Adds support for importing metadata directly from a CSV file. The file must be UTF-8.
- Changes indexing to batch items per collection. This amounts to a 9x speedup on importing large collections. This impacts this new csv import, the existing google import, and the rake task for reindexing all collections.

### DIFF
--- a/app/assets/javascripts/components/ImportResultsDialog.jsx
+++ b/app/assets/javascripts/components/ImportResultsDialog.jsx
@@ -1,0 +1,136 @@
+var React = require('react');
+var mui = require("material-ui");
+var Dialog = mui.Dialog;
+var RaisedButton = mui.RaisedButton;
+var FlatButton = mui.FlatButton;
+
+var ImportResultsDialog = React.createClass({
+  mixins: [MuiThemeMixin],
+
+  propTypes: {
+    open: React.PropTypes.bool.isRequired,
+    results: React.PropTypes.object.isRequired,
+  },
+
+  getInitialState: function() {
+    return {
+      open: this.props.open,
+    };
+  },
+
+  componentWillReceiveProps: function(nextProps) {
+    if(nextProps.open) {
+      this.setState({
+        open: nextProps.open,
+      });
+    }
+  },
+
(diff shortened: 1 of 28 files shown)


app/models/waggle/adapters/solr/session.rb
```diff
@@ -13,7 +13,7 @@ def connection
         end
 
         def index(*objects)
-          connection.add(*objects_as_solr(objects))
+          connection.add(objects_as_solr(*objects))
```
jonhartzler: I am curious.  What is the reason that this needed to be changed?

app/assets/javascripts/components/ImportResultsDialog.jsx
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    return [totals, newCount, changedCount, unchangedCount];
+  },
+
+  errors: function() {
+
+    var summary = this.props.results.summary;
+    if (summary.error_count > 0) {
+      var errors = this.props.results.errors;
```
hbeachey: Should this be props.results.summary.errors?

app/models/waggle/adapters/solr/session.rb
jgondron: Wasn't written correctly for allowing passing of arrays all the way through. Never had an issue before because even when we do reindex_all, it was looping through individual items and indexing.

app/assets/javascripts/components/ImportResultsDialog.jsx
jgondron: No, summary just has counts such as error_counts. errors has detailed validation errors for each row that failed to import. Ex:
`{
    "summary": {
        "total_count": 153,
        "valid_count": 152,
        "new_count": 7,
        "error_count": 1,
        "changed_count": 0,
        "unchanged_count": 145
    },
    "errors": {
        "0": {
            "errors": [
                "Year must be greater than 0 ",
                "Date published Invalid Date"
            ],
            "item": { ...
            }
        }
    }
}`

PR #469, opened 2016-11-16
Bugfix: Video thumbnail not showing correctly on Beehive pprd

jgondron:
The problem was duplication of items in the index, which caused react to flatten items in the array due to duplicate keys. We were using a combination of item.unique_id and item.type to uniquely identify the object to solr. We now have items with different types (ex: MetadataOnlyObject, VideoObject, etc), so this no longer works. Additionally, item.unique_id is only guaranteed to be unique within a collection. Changed this to use the collection.unique_id + item.unique_id when identifying the object in solr. This will mean we need to reindex after a deploy.

### DIFF
--- a/app/models/waggle/adapters/solr/index/item.rb
+++ b/app/models/waggle/adapters/solr/index/item.rb
@@ -11,7 +11,7 @@ module Waggle
           end
 
           def id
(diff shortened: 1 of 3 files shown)


jgondron:
This will require a reindex of everything. It should be much faster now with the changes to indexing.

PR #470, opened 2016-11-16
Fix showcase and page quick delete buttons

hbeachey:
The showcase and page quick delete buttons didn't check if the item could be deleted, which would break the site path. This no longer happens.

### DIFF
--- a/app/queries/item_query.rb
+++ b/app/queries/item_query.rb
@@ -27,11 +27,7 @@ class ItemQuery
     relation.find_by!(unique_id: id)
   end
 
-  def self.can_delete_item?(item)
-    !item.showcases.any? && !item.pages.any? && !item.children.any?
-  end
-
   def can_delete?
-    ItemQuery.can_delete_item?(relation)
+    CanDelete.item?(relation)
   end
 end
--- a/app/queries/page_query.rb
+++ b/app/queries/page_query.rb
@@ -23,11 +23,7 @@ class PageQuery
     relation.find_by!(unique_id: id)
   end
 
-  def self.can_delete_page?(page)
-    !SiteObjectsQuery.new.exists?(collection_object: page)
-  end
-
   def can_delete?
-    PageQuery.can_delete_page?(relation)
+    CanDelete.page?(relation)
   end
 end
--- a/app/queries/page_query.rb
+++ b/app/queries/page_query.rb
@@ -23,7 +23,11 @@ class PageQuery
     relation.find_by!(unique_id: id)
   end
 
+  def self.can_delete_page?(page)
+    !SiteObjectsQuery.new.exists?(collection_object: page)
+  end
+
   def can_delete?
-    !SiteObjectsQuery.new.exists?(collection_object: relation)
+    PageQuery.can_delete_page?(relation)
   end
 end
--- a/app/queries/showcase_query.rb
+++ b/app/queries/showcase_query.rb
@@ -27,11 +27,7 @@ class ShowcaseQuery
     relation.find_by!(unique_id: id)
   end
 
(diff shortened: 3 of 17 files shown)


app/queries/page_query.rb
```diff
@@ -23,7 +23,11 @@ def public_find(id)
     relation.find_by!(unique_id: id)
   end
 
+  def self.can_delete_page?(page)
```
jonhartzler: I would be tempted to move this to a policy object.    It is not really logic about the query.

app/queries/showcase_query.rb
```diff
@@ -27,7 +27,11 @@ def public_find(id)
     relation.find_by!(unique_id: id)
   end
 
+  def self.can_delete_showcase?(showcase)
```
jonhartzler: Same as above.

jgondron:
General question on all of the Destroy services: what should the behavior be when it cannot destroy? In the past we've allowed it to fall through to FKC checks at the data layer, which would throw an exception. Since we're handling checks at the app layer to see if there is an association in site_objects, I wonder if it shouldn't behave the same. Ie, if we get to the point where its actually trying to Destroy an object and it fails a CanDelete? check, raise an exception.

app/queries/item_query.rb
```diff
@@ -27,11 +27,7 @@ def public_find(id)
     relation.find_by!(unique_id: id)
   end
 
-  def self.can_delete_item?(item)
-    !item.showcases.any? && !item.pages.any? && !item.children.any?
-  end
-
   def can_delete?
```
jgondron: Should the #can_delete? methods even exist on the query objects any longer? Seems like we should we just migrate any users of these methods over to directly using the CanDelete service?

spec/services/destroy/item_spec.rb
```diff
@@ -76,14 +76,16 @@
       # it out will cause it to never call cascade on the child objects. Instead,
       # it tests to see that it calls destroy on the item's children, and it's
       # childrens' children, implying that cascade was properly called.
-      expect(child).to receive(:destroy!).twice
-      expect(child2).to receive(:destroy!).twice
+      allow(CanDelete).to receive(:item?)
```
jgondron: This should be separated into two specs. One that still checks that it actually calls item destroy, another that tests that it uses the CanDelete service before destroying

spec/services/destroy/item_spec.rb
```diff
@@ -76,14 +76,16 @@
...
+      expect(CanDelete).to receive(:item?).with(child).twice
+      expect(CanDelete).to receive(:item?).with(child2).twice
       subject.cascade!(item: item)
     end
 
     it "destroys sections before children to prevent FK constraints on Item->Section" do
       expect(destroy_section).to receive(:cascade!).with(section: section).at_least(:once).ordered
-      expect(child).to receive(:destroy!).at_least(:once).ordered
+      allow(CanDelete).to receive(:item?)
```
jgondron: Same as above

spec/services/destroy/item_spec.rb
hbeachey: Ideally yes, but since we don't actually destroy anything the children won't be removed from the parents and thus the destroy will only be called once for the bottom-most child as all the rest will fail the CanDelete check (they'll still have children).

hbeachey:
@jgondron I don't know that the FKC have actually worked for these cases in the past. I had to change how these delete utilities work because I was able to delete things that should not have been deletable and which, when deleted, broke the site path. I agree that we should stay consistent, but I also think changing these to be exceptions will be a considerable change and should be put into it's own task.

PR #480, opened 2017-01-03
DEC-1245 DEC-1246 DEC-1257 adding parent/children routes to items

hbeachey:


spec/decorators/v1/item_json_decorator_spec.rb
```diff
@@ -40,6 +40,22 @@
     end
   end
 
+  describe "#children" do
+    let(:item) { instance_double(Item, unique_id: "adsf", name: "name", children: ["a"]) }
+
```
jgondron: Add a test case for when the item has no children

spec/decorators/v1/item_json_decorator_spec.rb
```diff
@@ -40,6 +40,22 @@
...
+    it "returns the path to the children" do
+      expect(subject.children_url).to eq("http://test.host/v1/items/adsf/children")
+    end
+  end
+
+  describe "#parent" do
+    let(:item) { instance_double(Item, unique_id: "adsf", name: "name", parent: "a") }
+
```
jgondron: Add a test case for when an item has no parent

app/controllers/v1/items_controller.rb
```diff
@@ -56,6 +56,28 @@ def destroy
       end
     end
 
+    # show the items parent
+    def parent
```
jgondron: I don't think we actually need this one. The isPartOf/Parent url can just point to the show method for the parent object.

app/decorators/v1/item_json_decorator.rb
```diff
@@ -10,10 +10,22 @@ def at_id
       h.v1_item_url(object.unique_id)
     end
 
+    def parent_url
+      if object.parent
+        h.v1_item_parent_url(object.unique_id)
```
jgondron: Continuing from comment above, you can probably just use h.v1_item_url(object.parent.unique_id) instead of using this additional route.

spec/controllers/v1/items_controller_spec.rb
```diff
@@ -4,7 +4,7 @@
 RSpec.describe V1::ItemsController, type: :controller do
```
jgondron: Add test cases for the new methods

PR #482, opened 2017-01-05
DEC-1340 Indexing

hbeachey:
Altered indexing to correctly account for children. 

This required updating the rsolr gem from 1.0 to 2.0, which had a bug with only having one child, which I fixed in a fork of the repo.

I'm open to suggestions on what we want the identifier to be for parents. Right now it's just a string relation is_parent: "true" or "false"

### DIFF
--- a/app/models/waggle/adapters/solr/index/item.rb
+++ b/app/models/waggle/adapters/solr/index/item.rb
@@ -25,11 +25,23 @@ module Waggle
               hash.merge!(string_fields)
               hash.merge!(datetime_fields)
             end
+            if waggle_item.children && !waggle_item.children.empty?
+              @as_solr[:_childDocuments_] = children_field
+            end
             @as_solr[:last_updated_sort] = datetime_as_solr(waggle_item.send(:last_updated))
             @as_solr
           end
 
           private
+
+          def children_field
+            children = []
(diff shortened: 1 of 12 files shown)


app/models/waggle/item.rb
```diff
@@ -35,6 +40,20 @@ def type
       end
     end
 
+    def children
+      return [] if !decoratedItem
+      decoratedItem.children
+    end
+
+    def is_parent
```
hbeachey: This is not as simple of a change, we can change it here, but the method names is what the indexer is using to name its fields. 

See waggle/../solr/index/item.rb method `string_fields` to see how this is used.

Potential changes relate to initial comment of how we want to identify parents.

app/models/waggle/adapters/solr/index/item.rb
```diff
@@ -25,20 +25,33 @@ def as_solr
...
           def string_fields
             {}.tap do |hash|
               [
                 :at_id,
                 :unique_id,
                 :collection_id,
                 :type,
-                :thumbnail_url
+                :thumbnail_url,
+                :is_parent,
```
jgondron: Instead of is_parent: true|false, add "isPartOf/Parent": url to api show for the parent item

PR #483, opened 2017-01-10
Import parent/child associations

jgondron:
This changes the google and csv import routines so that they can handle parent/child associations identified by a "Parent Identifier" column.

Summary of changes:
- Changed the metadata rewrite to lookup the parent by the user defined id anytime a parent is given. If the parent exists, it adds the associated parent id to the child to create the association.
- Changed any methods that call CreateItems to give parents first, then children, to make sure that new parents are created before the lookup for the child is executed.
- Since the order that the cells are read is no longer preserved when passing the data on to the lower level CreateItems service, and this service is what adds any errors encountered (including the row where it encountered it), I had to start passing the original index down. So, changed CreateItems to expect an array of objects that have the original index along with the item properties.
- If the user puts a Parent Identifier column with a null value, the expected behavior is that it will clear any existing association for that item. But, the GoogleSession#worksheet_to_hash was only extracting columns that had values. Had to change this to make an explicit exception for Parent Identifier so that if it exists in the spreadsheet, it will always pass that along to the underlying classes in order to clear the association.

### DIFF
--- a/app/services/google_create_items.rb
+++ b/app/services/google_create_items.rb
@@ -52,9 +52,9 @@ class GoogleCreateItems
     children = []
     items.each.with_index do |item, index|
       if item["Parent Identifier"].present?
-        parents << { original_index: index, item_hash: item }
+        parents << { index: index, item_hash: item }
       else
-        children << { original_index: index, item_hash: item }
+        children << { index: index, item_hash: item }
       end
     end
     [parents, children]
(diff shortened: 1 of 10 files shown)


app/services/google_create_items.rb
```diff
@@ -46,6 +47,19 @@ def create_from_worksheet!(collection:, file:, sheet:)
 
   private
 
+  def parents_then_children(items:)
```
hbeachey: This looks good, I'd want to have it in a common module between the two import sections to be DRY.

PR #485, opened 2017-01-11
Dec 1254 child search

hbeachey:
Add the `../search/children` route to search only the children items of a collection.

By default, the resulting hits are returned grouped by the parent id. This can be turned off to return flat hits by adding the parameter `no_group=true` to the url.

### DIFF
--- a/app/controllers/v1/search_controller.rb
+++ b/app/controllers/v1/search_controller.rb
@@ -32,9 +32,7 @@ module V1
     private
 
     def collection
-      if params[:collection_id]
-        CollectionQuery.new.any_find(params[:collection_id])
-      end
+      @collection ||= params[:collection_id] ? CollectionQuery.new.any_find(params[:collection_id]) : nil
     end
   end
 end
--- a/app/controllers/v1/search_controller.rb
+++ b/app/controllers/v1/search_controller.rb
@@ -13,6 +13,22 @@ module V1
       @search = Waggle.search(**search_arguments)
     end
 
+    def children
+      group_by = params[:no_group] ? nil : "part_parent_s"
+
+      search_arguments = {
+        q: "-part_parent_s:_is_parent_ " + (params[:q] ? "AND " + params[:q] : ""),
+        facets: params[:facets],
+        sort: params[:sort],
+        rows: params[:rows],
+        start: params[:start],
+        filters: { collection_id: collection.unique_id },
+        group_by: group_by,
+        collection: collection,
+      }
+      @search = Waggle.search(**search_arguments)
+    end
+
     private
 
     def collection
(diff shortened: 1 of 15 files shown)


spec/models/waggle/adapters/solr/search/result_spec.rb
```diff
@@ -71,6 +71,39 @@
...
+        },
+      ).and_return("solr_response")
+      expect(subject.send(:connection)).to receive(:paginate).with(
+        2,
+        0,
+        "select",
+        params: {
+          q: "-part_parent_s:_is_parent_ AND a query",
+          group: true,
+          :"group.field" => "part_parent_s",
```
hbeachey: As far as I know, can't use this syntax with string keys. I was following the conventions set previously in these files.

app/controllers/v1/search_controller.rb
```diff
@@ -13,6 +13,22 @@ def index
       @search = Waggle.search(**search_arguments)
     end
 
+    def children
```
jgondron: Needs specs

app/models/waggle/adapters/solr/search/result.rb
```diff
@@ -13,6 +13,10 @@ def hits
             @hits ||= solr_docs.map { |solr_doc| Waggle::Adapters::Solr::Search::Hit.new(solr_doc) }
           end
 
+          def groups
```
jgondron: Needs specs

app/views/v1/search/_group.json.jbuilder
```diff
@@ -0,0 +1,4 @@
+json.groupId group.fetch(:id, "")
```
jgondron: The schema here could use a little work to look more like the rest of the objects. Right now, this produces:
```
{
  "groupId": "http://localhost:3017/v1/items/8e6a38769f",
  "hits": [
    {
      "@type": "SearchHit",
      "@id": "http://localhost:3017/v1/items/fad922ca22",
      "isPartOf/Item": "http://localhost:3017/v1/items/8e6a38769f",
      "type": "Metadata Only",
      "name": "CST Paragraph Level",
      "shortDescription": " ",
      "description": null,
      "thumbnailURL": null,
      "updated": "2017-01-11T18:51:51.000+00:00"
    }
  ]
}
```
Should perhaps be something more like:
```
{
  "@type": "SearchHitParent",
  "@id": "http://localhost:3017/v1/items/8e6a38769f",
  "children": [
    {
      "@type": "SearchHit",
      "@id": "http://localhost:3017/v1/items/fad922ca22",
      "isPartOf/Item": "http://localhost:3017/v1/items/8e6a38769f",
      "type": "Metadata Only",
      "name": "CST Paragraph Level",
      "shortDescription": " ",
      "description": null,
      "thumbnailURL": null,
      "updated": "2017-01-11T18:51:51.000+00:00"
    }
  ]
}
```

jonhartzler: Also consider using hits here so there is a consistent interface to the data.

app/controllers/v1/search_controller.rb
hbeachey: added

PR #488, opened 2017-01-20
Meta short description

jgondron:
Changed short_description to be pulled from the item instead of making up a value

### DIFF
--- a/app/models/waggle/search/hit.rb
+++ b/app/models/waggle/search/hit.rb
@@ -4,16 +4,11 @@ module Waggle
       attr_reader :adapter_hit
       private :adapter_hit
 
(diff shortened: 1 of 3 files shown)


app/models/waggle/search/hit.rb
```diff
@@ -4,16 +4,11 @@ class Hit
...
 
-      delegate :name, :at_id, :part_parent, :type, :description, :date_created, :creator, :thumbnail_url, :last_updated, to: :adapter_hit
+      delegate :name, :at_id, :part_parent, :type, :description, :short_description, :date_created, :creator, :thumbnail_url, :last_updated, to: :adapter_hit
 
       def initialize(adapter_hit)
         @adapter_hit = adapter_hit
       end
-
-      def short_description
-        "#{creator} #{date_created}"
```
hbeachey: Hopefully nothing was depending on this. Beehive?

PR #490, opened 2017-03-13
Clear collection

jgondron:
Created a rake task to clear out a collection. Eventually it might make sense to have this as an actual service that can be reached from some admin function on a collection, but I didn't want to touch app code for this since we're limited on time.


rfox2:
I approve retrospectively

PR #495, opened 2017-09-29
React fixes

danthewolfe:
* Update React and Material UI
* Fix warnings related to depreciations and lack of keys.


rfox2:
Should the travis errors be addressed before we merge? Is that a build issue on travis?

PR #515, opened 2020-01-24
Few fixes for the adhoc import task

jgondron:
Something weird happens with openuri#open nowadays when it parses the filename that seems to kill the honeypot job. Changed to URI.parse and it preserves the original file name, which seems to work better.

Changed to use Metadata::Setter for setting the item name based on file name. (see https://github.com/ndlib/honeycomb/pull/326)


jgondron:
At the moment, we are pulling these from http, so that's all I can confirm works. We may have to change again if we find we need to pull from a mounted fs.

PR #521, opened 2020-03-30
DEC-847 Fix invalid dates importing from csv without error.

mlkirkpatrick:


danthewolfe:
Check Travis failures.

mlkirkpatrick:
Yeah, idk what that's about. Looks like it has been failing for several months now. I fixed one immediately obvious issue, but not sure why so many unit tests are failing. In any case, I'm pretty sure it's not related to this PR, but it does need to be addressed regardless.

danthewolfe:
@mlkirkpatrick That's why I still gave it a pass, failures didn't look related but worth checking.

PR #535, opened 2020-05-11
Allow faceting on other field types besides strings.

mlkirkpatrick:


mlkirkpatrick:
POs decided not to do change this.

PR #543, opened 2020-08-05
Create dockerized environment

ialford:
This adds all the files and changes to create a rudimentary development environment within Docker.

Docker will create three containers: one for Postgres, one for Solr, and one for Ruby.

Instructions are also included in the `docker/` folder.

### DIFF
--- a/docker/rails_entry.sh
+++ b/docker/rails_entry.sh
@@ -1,0 +1,23 @@
+# All the things that will execute when starting the rails service
+# Copy the generated lock file back into the project root. This is to sync this change back to the
+# application after the container has mounted the sync volume
+
+cp /bundle/Gemfile.lock ./
+
+# Have to do this here to allow using localhost when deployed with ECS awsvpc
+# Could not see a way to have this solr.yml read env
+# sed -i "s;\${SOLR_URL};$SOLR_URL;g" /project_root/config/solr.yml
+
+### Wait for dependencies
+chmod u+x docker/wait-for-it.sh && bash docker/wait-for-it.sh -t 120 ${DB_HOST}:5432
+bash docker/wait-for-it.sh -t 120 ${SOLR_HOST}:8983
+
+# In production we'll be using a persistant database, but not a solr instance,
+# so reindex everything on start
+# bundle exec rake solr:index:all
+
+### Uncomment below for development ###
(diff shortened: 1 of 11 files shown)


docker/rails_entry.sh
```diff
@@ -0,0 +1,24 @@
...
+# sed -i "s;\${SOLR_URL};$SOLR_URL;g" /project_root/config/solr.yml
+
+### Wait for dependencies
+chmod u+x docker/wait-for-it.sh && bash docker/wait-for-it.sh -t 120 ${DB_HOST}:5432
+bash docker/wait-for-it.sh -t 120 ${SOLR_HOST}:8983
+
+bundle exec rake assets:precompile
+
+### Run setup on the fresh database
+bundle exec rake db:setup RAILS_ENV=development
```
jgondron, line 18: Note: This is fine for now in a dev environment, but you will definitely not want to do this once we connect it to an external database.

docker/rails_entry.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# Have to do this here to allow using localhost when deployed with ECS awsvpc
+# Could not see a way to have this solr.yml read env
+# sed -i "s;\${SOLR_URL};$SOLR_URL;g" /project_root/config/solr.yml
+
+### Wait for dependencies
+chmod u+x docker/wait-for-it.sh && bash docker/wait-for-it.sh -t 120 ${DB_HOST}:5432
+bash docker/wait-for-it.sh -t 120 ${SOLR_HOST}:8983
+
+bundle exec rake assets:precompile
```
jgondron, line 15: Are you sure you want to do this every time you start rails? This should only really need to be done once at build time. In production, we'll want this built into the image otherwise it will take much longer to start and you might timeout on healthchecks etc

ialford: You're right. I'll move to the Dockerfile.

docker/rails_entry.sh
ialford: Will remove and add instructions for README

ialford: Updated

docker/rails_entry.sh
ialford: Moved to README

PR #545, opened 2020-09-22
nginx docker

ialford:
adding in nginx container to be able to run separate services, similar to our other services

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -1,0 +1,23 @@
+# To also start up the environment as a background daemon: docker-compose up -d
+# Both of the above commands run from the root of the repository
+version: '3.2'
+services:
+  rails:
+    environment:
+      RAILS_ENV: production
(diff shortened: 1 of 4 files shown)


rdoughty (approved):

rdoughty:
Looks like Travis is failing - The command "bundle exec rspec spec" exited with 1.

ialford:
> @ialford looks like Travis is failing - The command "bundle exec rspec spec" exited with 1.

@rdoughty yes, I have seen that. There's one unit test failing - and @rfox2 has said it's a known issue. It should probably be addressed at some point.

mlkirkpatrick (approved):

mlkirkpatrick:
Just an FYI about the unit tests in CI build: It seems to be a race condition. When I run unit tests locally they all succeed (presumably because they are running in the same thread/synchronously?) Fixing CI should just be a matter of updating the tests so no 2 tests rely on the same user id, or if there is a way to configure it to run differently that might work too.

PR #548, opened 2020-11-05
Add patch to account for BC dates on item records

rfox2:
Dates that are flagged BC on the item edit screen were not being indexed. This was happening because waggle was not set up to correctly parse BC dates which have a prepended "-" to indicate "negative" I'm assuming. The waggle method needed to have a conditional added that would account for BC dates so that dates with no month or day designation get correctly formatted and parsed prior to indexing.


.gitignore
```diff
@@ -5,6 +5,12 @@
 # or operating system, you probably want to add a global ignore instead:
 #   git config --global core.excludesfile '~/.gitignore_global'
 
+# Ignore local scripts
+/bin/*
+
+# Ignore development env file
+/config/environments/development.rb
+
```
mlkirkpatrick, line 13: Why exclude these? Doesn't that make it difficult to just clone the repo and run the app?

If the config doesn't work as is, why don't we change it?

rfox2: We can - I was just hesitant because the old port number doesn't work with Okta (you need 3000 for any app you want to test in development). I am perfectly happy adding those changes in.

rfox2: Changed back but with port 3000 so your mileage may vary...

mlkirkpatrick (approved):

rdoughty (approved):

jonhartzler (approved):

PR #552, opened 2021-02-19
More containerization

jgondron:
Application changes

- Changed to allow configuring logging via env
- Changed production log level to allow override through env, with :warn
as the default. Previous was :info, which was way to verbose for
production. Also removed the colorizing in the logs in production since
we likely won't be reading these logs via the shell.
- Changed the secrets around a bit to mostly read from env as the example.
Added read from env for the Google secrets and added a mapping for these
in the compose. There's still something not quite right with the post
to export_google_sheet_callback. The gem may need to be updated, which
may require application changes.
- Changed the solr config to support reading env. The previous method of
replacing config values at rails start was not conducive to running rake
tasks that also read solr config from env.
- Changed the devise initializer to skip configuration when we are
precompiling assets. As far as I can tell there's no way to determine this
state from within a config natively in rails, so when precompiling we need
to now add ASSET_PRECOMPILE=true to the env, otherwise we'll get a missing
secrets exception.
- Change AWS security chain handling for creating s3 signed url
- Fixed a client header issue when exporting to Google. Appears to have
been necessary due to a change to the Google APIs that had been made
since we last deployed Honeycomb.
- Changed the sneakers config to read in the necessary secrets from
env and to allow logging to stdout when in a container env.

Docker changes

- Added a static SECRET_KEY_BASE to the compose production file so that
the service would fully start in production. This value will different
when deployed. It's only here to run the containers in production mode
locally to test
- Added exit handling to the wait-for-it calls in the rails entry. The
service was failing to connect to the database, but stabilizing anyway
because it was not exiting.
- Due to a [change in Fargate 1.4](https://github.com/aws/containers-roadmap/issues/863),
mounting the precompiled asset volume had to be changed to use a VOLUME
instruction inside of Dockerfile.rails and the nginx container changed
to use volumes_from. Unfortunately, the [volumes_from directive is not available in compose 3](https://docs.docker.com/compose/compose-file/compose-versioning/#version-3).
In order to match how this will be done out in ECS, I changed the
compose version to 2.
- Added a volume to share static images with nginx container.
- This also changes the rails container to share the entire /public dir
as a volume and sym link the images to /mnt/honeycomb. This is so that
we can mount efs to /mnt/honeycomb and still share /public as a volume
to nginx. Nginx will now try to serve the request from /public and fall
back to the upstream rails app if it doesn't find a static file.
- Added an "alb" endpoint in prod compose to terminate ssl when running
locally as production. This is for the okta login/callback to work when
run in a compose env


docker-compose.yml
```diff
@@ -11,31 +11,39 @@ services:
...
       SOLR_PORT: 8983
       OKTA_CLIENT_ID: ${OKTA_CLIENT_ID}
       OKTA_CLIENT_SECRET: ${OKTA_CLIENT_SECRET}
-      OKTA_REDIRECT_URL: ${OKTA_REDIRECT_URL}
-      OKTA_BASE_URL: ${OKTA_BASE_URL}
-      OKTA_LOGOUT_URL: ${OKTA_LOGOUT_URL}
       OKTA_AUTH_ID: ${OKTA_AUTH_ID}
+      OKTA_REDIRECT_URL: https://localhost:3000/users/auth/oktaoauth/callback
+      OKTA_BASE_URL: https://okta.nd.edu/oauth2/
+      OKTA_LOGOUT_URL: https://okta.nd.edu/
```
mlkirkpatrick, line 33: Why hardcode the Okta values? It probably doesn't matter as they aren't likely to change, but it looks like you're still using the environment variables in secrets. So I'd think it would be safer to just use those so you don't end up with a discrepancy.

mlkirkpatrick (approved):

app/workers/retry_worker.rb
```diff
@@ -25,10 +25,10 @@ def self.number_of_workers
   def work(*args)
     original_work(*args)
   rescue StandardError => e
-    Raven.capture_exception(e)
     logger.error e.message
     logger.error args
     logger.error e.backtrace.join("\n")
+    Raven.capture_exception(e)
```
rfox2, line 31: Have you tested this? I'm not sure that we want the formatting in the Sentry message.

rfox2: Perhaps those calls to the error are non-mutable / non-destructive, in which case this doesn't matter

app/models/waggle/adapters/solr/session.rb
```diff
@@ -54,7 +54,10 @@ def solr_objects(*objects)
         end
 
         def load_configuration
-          YAML.load_file(File.join(Rails.root, "config", "solr.yml")).fetch(Rails.env)
+          path = File.join(Rails.root, "config", "solr.yml")
+          template = ERB.new(File.read(path))
```
rfox2, line 58: Why is this necessary?

rfox2: I may have answered my question. It looks like you wanted to use the marshaled data structures later on.

rfox2 (changes requested):
Not really requesting changes - just wanted some questions answered. I'm good with all of the changes as a whole.

app/workers/retry_worker.rb
jgondron: This shouldn't do any formatting of the Sentry message. This is just a change to get it to output to the log first. I had hit some exceptions in my environment that were hard to troubleshoot because the Raven call was also failing, so I'd get no output.

docker-compose.yml
jgondron: This is assigning the environment variables so that they're picked up by secrets when running locally. This is just to limit the amount of ENV setup a developer has to do when running the compose locally

app/models/waggle/adapters/solr/session.rb
jgondron: It's so that it processes calls to assign these from ENV, ex: https://github.com/ndlib/honeycomb/pull/552/files#diff-8e034ea11678bcb88ffb7b2dc0bb7adb9ce10b62471581a2a0c7851de8420011R16-R17. YAML.load_file does not do this for us, have to use ERB

app/workers/retry_worker.rb
rfox2: Awesome thanks

app/models/waggle/adapters/solr/session.rb
rfox2: Gotcha

rfox2 (approved):

PR #555, opened 2021-03-11
Add basic smoke tests

jgondron:
This primarily adds tests that are a bit more basic than the others for
running smoke tests. Additionally includes some refactoring of env keys
to be more consistent with other services in the stack.

### DIFF
--- a/docker/rails_entry.sh
+++ b/docker/rails_entry.sh
@@ -2,7 +2,7 @@
 ln -s /mnt/honeycomb /project_root/public/system/images
 
 ### Wait for dependencies
-if ! docker/wait-for-it.sh -t 120 ${DB_HOST}:5432; then exit 1; fi
+if ! docker/wait-for-it.sh -t 120 ${DB_HOSTNAME}:5432; then exit 1; fi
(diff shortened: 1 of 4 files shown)


docker/rails_entry.sh
```diff
@@ -2,7 +2,7 @@
 ln -s /mnt/honeycomb /project_root/public/system/images
 
 ### Wait for dependencies
-if ! docker/wait-for-it.sh -t 120 ${DB_HOST}:5432; then exit 1; fi
+if ! docker/wait-for-it.sh -t 120 ${DB_HOSTNAME}:5432; then exit 1; fi
 if ! docker/wait-for-it.sh -t 120 ${SOLR_HOST}:8983; then exit 1; fi
```
mlkirkpatrick, line 6: Should these use DB_PORT and SOLR_PORT?

mlkirkpatrick (approved):

ialford (approved):

wsill (approved):

docker/rails_entry.sh
jgondron: Nah, it's hostname:port. This is using DB_HOSTNAME from env as the hostname, and a constant value for the port. Similar for solr.

PR #562, opened 2021-06-08
Fix editors in Docker env

jgondron:
The previous deploy strategy expected a config/hesburgh_api.yml to exist
on the host. This file configures how to connect to the Hesburgh API for
searching for users. This PR is to add this file to the Docker build
and instructs rails to read the url and token values from env. This should
fix an issue with searching for users when adding admins to the site, or
when adding editors to a collection.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -18,7 +18,7 @@ GIT
 
 GIT
   remote: https://github.com/ndlib/hesburgh_api.git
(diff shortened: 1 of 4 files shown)


mlkirkpatrick (approved):

ialford (commented):
Do you also need to change the SHA at https://github.com/ndlib/honeycomb/blob/master/Gemfile.lock#L21 to pull the latest gem updates?

ialford (approved):

